PR #1, opened 2020-08-14
Move R part of generateModuleInstallingR() to common

TimKDJ:
Also adds a function that takes a description file and spits out the dependencies.

Ok so basically I just moved the R part from the cpp function `generateModuleInstallingR()`, but I was a little confused at its setup (and I'm curious if some of it could be simplified by overwriting system2, see [this](https://r.789695.n4.nabble.com/Capture-output-of-install-packages-pipe-system2-td4676754.html) and [this](https://stackoverflow.com/questions/12416076/suppress-install-outputs-in-r)). But I didn't want to go too deep into it. If this is not what you had intended, we can zoom @JorisGoosen.

Mapping of R to C++ args:
```
modulePkg      = _modulePackage
libPathsToUse  = getLibPathsToUse()
moduleLibrary  = moduleRLibrary().toStdString()
repos          = Settings::value(Settings::CRAN_REPO_URL).toString().toStdString()
onlyModPkg     = onlyModPkg
```

Possible issues
1. jaspModules can exist in multiple folders (if you `install_deps` and one of the deps is a jaspModule it will also install that, unless it's found in one of the other libraries)
2. Unless the environment variable `GITHUB_PAT` is set, the `remotes` package uses a default public Github PAT which has a limit of 60 queries per IP address per hour.

See https://github.com/jasp-stats/INTERNAL-jasp/issues/1014#issuecomment-673405667

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -1,0 +1,2 @@
+^.*\.Rproj$
+^\.Rproj\.user$
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,49 @@
+# History files
+.Rhistory
+.Rapp.history
+
+# Session Data files
+.RData
+
+# User-specific files
+.Ruserdata
+
+# Example code in package build process
+*-Ex.R
+
+# Output files from R CMD build
+/*.tar.gz
+
+# Output files from R CMD check
+/*.Rcheck/
+
+# RStudio files
+.Rproj.user/
+
+# produced vignettes
+vignettes/*.html
+vignettes/*.pdf
+
+# OAuth2 token, see https://github.com/hadley/httr/releases/tag/v0.3
+.httr-oauth
+
+# knitr and R markdown default cache directories
+*_cache/
+/cache/
+
+# Temporary files created by R markdown
+*.utf8.md
+*.knit.md
+
+# R Environment Variables
+.Renviron
+
+# pkgdown site
+docs/
+
+# translation temp files
+po/*~
+
+# OSX files
+*.DS_Store
+Thumbs.db
--- a/R/common.R
+++ b/R/common.R
@@ -263,7 +263,7 @@ initEnvironment <- function() {
 
   if (identical(.Platform$OS.type, "windows"))
     assignFunctionInPackage(fakeGrDevicesPdf, "pdf", "grDevices") # this fixes the problem that grDevices::pdf() does not work within JASP (https://github.com/jasp-stats/INTERNAL-jasp/issues/682)
-  
+
   for (package in packages)
     if (base::isNamespaceLoaded(package) == FALSE)
       try(base::loadNamespace(package), silent=TRUE)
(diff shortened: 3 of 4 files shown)


TimKDJ:
And I guess `pkgName` is actually not required, because we can go obtain that from `devtools::as.package(modulePkg)`. But what sort of formats can modulePkg be?

TimKDJ:
I doubt this is still useful / would need to be redone to be useful
Can I close it @JorisGoosen?

JorisGoosen:
Aah I forgot about this PR...

What changed to make you change your mind about it?
If this can actually handle particular versions and upgrades properly that would be an improvement.

If you want to close it that is fine btw!

JorisGoosen:
So I still think something like this PR is required and it might still be this code, depending on the answer to the following:
The question is why you felt it necessary to add `getDepsFromDescription`? Because R is quite capable of parsing DESCRIPTION right? Or is this what corresponds to [loadRequiredModulesFromDESCRIPTIONTxt](https://github.com/jasp-stats/jasp-desktop/blob/0ae466bb9434f528631a68349d9af1774b59b286/Desktop/modules/dynamicmodule.cpp#L183)?

TimKDJ:
I added it because of the first bullet of https://github.com/jasp-stats/INTERNAL-jasp/issues/1014#issuecomment-673405667
But your cpp solution works too. R doesn't do a whole lot of parsing in the functions they made available in their utils package (it just becomes a single string):

```
> packageDescription("jaspTTests")
Package: jaspTTests
Type: Package
Title: T-Tests Module for JASP
Version: 0.11.0
Date: 2019-08-31
Author: JASP Team
Website: jasp-stats.org
Maintainer: JASP Team <info@jasp-stats.org>
Description: This module offers the standard Frequentist and Bayesian T-Tests analyses.
License: GPL (>= 2)
Encoding: UTF-8
Imports: BayesFactor, BSDA, car, ggplot2, jaspBase, jaspGraphs, logspline, plotrix, plyr
Remotes: jasp-stats/jaspBase, jasp-stats/jaspGraphs
RemoteType: local
RemoteUrl: /Users/tdejong/Documents/GitHub/jaspTTests
NeedsCompilation: no
Packaged: 2020-11-12 11:02:20 UTC; tdejong
Built: R 3.6.3; ; 2020-11-12 11:02:21 UTC; unix

-- File: /Library/Frameworks/R.framework/Versions/3.6/Resources/library/jaspTTests/Meta/package.rds 
> a<-packageDescription("jaspTTests")
> a$Imports
[1] "BayesFactor, BSDA, car, ggplot2, jaspBase, jaspGraphs,\nlogspline, plotrix, plyr"
```

I could see if I can make this PR work though.

JorisGoosen:
Ill put your code in the repo. As it doesnt hit anything else it shouldn't be a problem to have it in there.
Then I can prepare a branch of jasp-desktop that uses it so that Don can work on `renv`.

I'll also be moving this code to a separate R file, as there is very little reason to keep pushing all code into `common.R`.

PR #5, opened 2020-11-13
export interaction symbol as replacement for .unvf

vandenman:
So while removing the `.v` calls from the Bayesian ANOVAs, I encountered one situation we haven't standardized yet: `.unvf` changes `:` into `✻`. To avoid that every analysis has to figure this out for themselves, we should just put this in jaspBase.

The implementation in this PR:

1. exports `interactionSymbol` with the appropriate encoding.
2. exports `substituteInteractionSymbol` that changes e.g., `"a:b"` into `"a ✻ b"`.


TimKDJ:
I don't suppose we can do this automatically after an analysis run, but before the columns are decoded?
It's a nice thing for this to be standardized across analyses, but it might be a bit tedious for developers to do this manually

vandenman:
So, you propose translating all `:` into `✻` automatically? Won't that be a problem when people want to show `:`? I suppose we could escape `:` somehow (e.g., `\\:`) though so that it's also possible that the encoding doesn't translate it.

@JorisGoosen would this be straightforward to implement?

TimKDJ:
Just whenever you find `variable:variable`

vandenman:
So a table like this still has to be done manually?

image

In that case, we might as well put this stuff in `jaspBase`. Right now, both `jaspRegression` and `jaspAnova` need this (unfortunately, they can't reuse code because BayesFactor and lm format the output differently).

TimKDJ:
And adding a new column type for model terms? (Basically I'm just looking for a way that would make it a little easier)

JorisGoosen:
You could just make a little R-function that replaces `:` by the fancy `*` right?
Columnnames will all be encoded anyway so you do not need to worry about it containing `:` same goes for levels, those are encoded as well.

So a simple replace-char function can be part of jaspBase I guess? It is also just a oneliner in R anyway no?

JorisGoosen:
Ps. I think it would be a bit confusing to have this be done "automatically" during the "decoding" to be honest.
And might lead to weird unexpected  behaviour whenever anyway decides they *do* want `:`

TimKDJ:
So that makes it 2 v 1, we'll do the manual replace then

NAMESPACE
```diff
@@ -28,3 +28,5 @@ export(.RCodeInOptionsIsOk)
 S3method(.RCodeInOptionsIsOk, list)
 S3method(.RCodeInOptionsIsOk, default)
 export(.parseAndStoreFormulaOptions)
+export(substituteInteractionSymbol)
+export(interactionSymbol)
```
TimKDJ, line 32: I think given the low relative importance of this functionality, just the export of the substitute function should be enough.

R/base64.R
```diff
@@ -28,3 +28,11 @@
   else
     decodeColNames(x, ...)
 }
+
+interactionSymbol <- "\u2009\u273B\u2009"
+base::Encoding(interactionSymbol) <- "UTF-8"
+
+substituteInteractionSymbol <- function(strings) {
```
TimKDJ, line 35: `prettyInteraction()` or `subInteractionSymbol()`? I'm all for descriptive names, but it feels like this could be a bit more succinct ;)

R/base64.R
```diff
@@ -28,3 +28,11 @@
...
+  # assumes the input consists of encoded column names, which never contain ':'
+  gsub(":", interactionSymbol, strings, fixed = TRUE)
```
TimKDJ, line 37: Suggestion: add functionality to go through a list recursively so people can just chuck in the model object returned from a package

TimKDJ (changes requested):

NAMESPACE
vandenman: Well, there are already quite a few `paste(strings, collapse = "\u2009\u273b\u2009")` in there (I count 5 analyses using it, 6 if logistic regression would actually use the nice symbol instead of `x`). Since we don't encode the factor levels in the data set (although maybe we should), the replacement function is kinda risky.

R/base64.R
vandenman: how about `gsubInteractionSym`?

TimKDJ: Sounds good!

NAMESPACE
TimKDJ: Hmm if it's used like that I guess it'd be nice to export. Just make sure the naming of the two exports match (as per the next comment)

`gsubInteractionSym(paste(strings, collapse = ":"))` seems a bit over the top :p

JorisGoosen: > Since we don't encode the factor levels in the data set (although maybe we should), the replacement function is kinda risky.

If we don't trust people to create sensible columnnames we can also not trust them to make sensible levels right?

TimKDJ: Kinda glossed over that, why don't we encode the levels?

vandenman: there was no reason to encode them... so far :P

vandenman: but when separating the column names from the factor levels in the R output (see example above) it could very well be that people use some weird symbols that mess up my regex.

JorisGoosen: So we should make two new issues I think, one requesting that levels are encoded along with the columnnames.
The other I was talking about with Don and that would be to also include those encodings in jaspTools so that the unittests test the exact same way as JASP.

vandenman: Here you go: https://github.com/jasp-stats/INTERNAL-jasp/issues/1171, https://github.com/jasp-stats/INTERNAL-jasp/issues/1172

NAMESPACE
```diff
@@ -28,3 +28,7 @@ export(.RCodeInOptionsIsOk)
 S3method(.RCodeInOptionsIsOk, list)
 S3method(.RCodeInOptionsIsOk, default)
 export(.parseAndStoreFormulaOptions)
+export(gsubInteractionSym)
+S3method(gsubInteractionSym, list)
+S3method(gsubInteractionSym, character)
+export(interactionSymbol)
```
TimKDJ, line 34: I think it's good to use either `sym` or `symbol` for both

PR #9, opened 2020-12-23
adding .getSeedJASP

FBartos:
this function was part of the base JASP functions but then disappeared - I was using it in Mixed Models and I started getting errors

also, there was an error check that was commented out from times when the .setSeed qml component wasn't implemented, I think that we can bring it back


R/common.R
```diff
@@ -2881,8 +2881,21 @@ editImage <- function(optionsJson) {
...
+  }
+}
+
+.getSeedJASP <- function(options) {
+  
+  if (is.list(options) && c("setSeed", "seed") %in% names(options)) {
+    if (isTRUE(options[["setSeed"]]))
+      return(options[["seed"]])
+  } else {
+    stop(paste(".getSeedJASP was called with an incorrect argument.",
```
TimKDJ, line 2897: I think this will pose a problem for functions that are shared between analyses (or even modules) when one analysis defines the seed component in the qml interface and the other does not. I don't know if such shared functions still exist, but we should be sure before uncommenting this.

R/common.R
```diff
@@ -2881,8 +2881,21 @@ editImage <- function(optionsJson) {
...
     if (isTRUE(options[["setSeed"]]))
       set.seed(options[["seed"]])
+  } else {
+    stop(paste(".setSeedJASP was called with an incorrect argument.",
+               "The argument options should be the options list from QML.",
+               "Ensure that the SetSeed{} QML component is present in the QML file for this analysis."))
+  }
+}
+
+.getSeedJASP <- function(options) {
```
TimKDJ, line 2891: Can you add an export to the `NAMESPACE`?

R/common.R
FBartos: oh, true. I forget that we have something like that in the t-tests :/ not great, I will comment it out again then

TimKDJ: Could you also add a comment explaining why the code is commented out? Will probably be useful in the future

TimKDJ: Note that there are two of these `stop()`'s

PR #13, opened 2021-01-27
Update package Rcpp 1.0.6

FransMeerhoff:
Needed for update Bayesrel v 0.7.0.3
Request Julius 26-01-2021


JorisGoosen (changes requested):
Shouldn't Bayesrel also be in here?
And maybe bain?

JorisGoosen (approved):

PR #15, opened 2021-02-10
add hasSubstring

JorisGoosen:


R/friendlyConstructorFunctions.R
```diff
@@ -111,3 +111,7 @@ zScores.factor <- function(x) {
   else # assume this is really what the user wants
     return(scale(x))
 }
+
+hasSubstring <- function(string, substring) {
+	return(stringi::stri_count(str=string, fixed=substring) > 0)
```
vandenman, line 116:
```suggestion
	return(grepl(pattern = substring, x = string, fixed = TRUE))
```

vandenman (changes requested):
see above, I'd prefer a base R solution

vandenman (approved):

PR #16, opened 2021-02-15
whitelist checker in r with parsing

vandenman:
Example for https://github.com/jasp-stats/INTERNAL-jasp/issues/808

Two things I'm not entirely happy with:

1. How to construct the whitelist? Hard coding it in two spots doesn't sound very nice to me. Note that this approach requires the whitelist to be encoded as functions rather than strings.
2. How to compare two functions? Right now I'm comparing R functions rather than strings and I think that R under the hood compares the pointers. This approach is linear in the length of the whitelist and we might be able to do better. Right now the approach is quite fast. When a function is not in the whitelist, the most common but also worst-case scenario, it takes about 4 ms to iterate through a list of 11000 functions (I'm praying our whitelist never becomes bigger than that).

### DIFF
--- a/R/applyWhiteList.R
+++ b/R/applyWhiteList.R
@@ -1,0 +1,112 @@
+
+rCodeValidator <- function(code) {
+  # this one should be called from JASP!
+  #' @param code a character string with R code
+
+  result <- tryCatch(
+    validateStringRCode(code),
+    invalidCodeError = function(e) { list(pass = "false", reason = e[["message"]])     },
+    error            = function(e) { list(pass = "false", reason = "unexpected error") }
+  )
+
+  return(toJSON(result))
+
+}
+
+fakeAssign <- function(expr, envir) {
+  assign(x = as.character(expr), value = NULL, envir = envir)
+}
+
+validateRawRCode <- function(rcode) {
+  expr <- substitute(rcode)
+  return(validateRExpression(expr))
+}
+
+validateStringRCode <- function(rcode) {
+  expr <- parse(text = rcode)
+  return(validateRExpression(expr))
+}
+
+validateRExpression <- function(expr) {
+  envir <- new.env(parent = baseenv())
+  validateInner(expr, envir)
+  return(list(pass = "true", reason = ""))
+}
+
+validateInner <- function(code, envir) {
+  for (i in seq_along(code)) {
+    line <- code[[i]]
+    validateAndEval(line, envir)
+  }
+}
+
+validateAndEval <- function(expr, envir) {
+
+  if (is.symbol(expr)) {
+
+    validateSymbol(expr, envir = envir)
+    return()
+
+  } else if (is.call(expr)) {
+
+    e <- try(print(expr))
+
+    if (is.assignment(expr[[1L]])) {
+
+      if (is.assignment(expr[[3L]]))
+        invalidCodeError(gettext("It's not allowed to reassign the assignment operator!"))
+
+      if (length(expr[[3L]]) > 1L)
+        validateInner(expr[[3L]], envir = envir)
+      else
+        validateAndEval(expr[[3L]], envir)
+
+      fakeAssign(expr[[2]], envir = envir)
+
+    } else {
+
+      validateInner(expr, envir = envir)
+
+    }
+  }
+}
+
+validateSymbol <- function(sym, envir) {
+
+  if (!is.symbol(sym))
+    return()
+
+  name <- as.character(sym)
+  val  <- mget(name, envir = envir, ifnotfound = NA, inherits = TRUE)[[1L]]
+  if (!is.function(val))
+    return()
+
+  if (!inWhiteList(val))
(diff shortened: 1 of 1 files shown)


JorisGoosen:
>     1. How to construct the whitelist? Hard coding it in two spots doesn't sound very nice to me. Note that this approach requires the whitelist to be encoded as functions rather than strings.

I could make the c++-whitelist available as a charvector and then we could look up the function by name?

>     2. How to compare two functions? Right now I'm comparing R functions rather than strings and I think that R under the hood compares the pointers. This approach is linear in the length of the whitelist and we might be able to do better. Right now the approach is quite fast. When a function is not in the whitelist, the most common but also worst-case scenario, it takes about 4 ms to iterate through a list of 11000 functions (I'm praying our whitelist never becomes bigger than that).

If it really compares pointers I think we could do that *much better* by using `std::set<void*>` or something.
Maybe that can hook into the problem under 1., we construct the set in c++ together with an R-helper-function or something to get the pointer that belongs to the functionname and then we add a "functionInWhitelist" `jaspRcpp` func?

vandenman:
> I could make the c++-whitelist available as a charvector and then we could look up the function by name?

So the problem with this is that I don't store the functions by symbol, but by value. Or rather, I just ask R, "what do you get for this symbol" and if the result is a function that's not in the whitelist, it's bad. We could also do the bookkeeping of assigned names ourselves, instead of having R resolve the name but then I need to adjust the code a bit.

> If it really compares pointers I think we could do that much better by using std::set<void*> or something.

I think the relevant line is this: https://github.com/wch/r-source/blob/4b24a12b68394e8772f50909d69fb428e9ec0844/src/main/identical.c#L91, but the entry point for R's `identical` is [here](https://github.com/wch/r-source/blob/4b24a12b68394e8772f50909d69fb428e9ec0844/src/main/identical.c#L42) and that code does a lot more than what's necessary. 

Either way, a C++ function that does the lookup with a set is probably best.

R/applyWhiteList.R
```diff
@@ -0,0 +1,112 @@
...
+  # TODO: figure out if we can do this a little bit more efficiently. Perhaps in c++?
+  for (fun in whiteList)
+    if (identical(f, fun))
+      return(TRUE)
+  return(FALSE)
+}
+
+# TODO: we need some way to fill this nicely... also we may want to prefixes the namespaces sometimes
+whiteList <- list(
+  list, `<-`, `{`, `[[`, print, `for`, `if`, `&&`, `||`, `c`, `+`, `-`, `/`, `*`, `:`, `>`, `body`, `function`, `c`
```
JorisGoosen, line 106: Aaaah, I was misled by the backticks, I thought I saw strings and didnt think (or look) harder after that...

JorisGoosen:
> > I could make the c++-whitelist available as a charvector and then we could look up the function by name?
> 
> So the problem with this is that I don't store the functions by symbol, but by value. Or rather, I just ask R, "what do you get for this symbol" and if the result is a function that's not in the whitelist, it's bad. We could also do the bookkeeping of assigned names ourselves, instead of having R resolve the name but then I need to adjust the code a bit.

Well, I think comparing to the "value" is a good choice, I figured it wouldn't be hard to get R to divulge what belongs to a given symbol. To get it working with that I could actually just use look up the value for each whitelisted function in R directly.
I'm pretty sure it is going to be easy to go from a symbol to a value, maybe as easy as calling "parse" or calling some function.

But I can add that later. I think putting this PR in jaspBase@master isn't necessary btw. It might be cleaner to simply open a new branch in jasp-stats/jaspBase and use that in the branch of jasp-desktop that I'll use to test this.
That way we can work on it without adding a whole bunch of commits to master here before it is done.

> 
> > If it really compares pointers I think we could do that much better by using std::set<void*> or something.
> 
> I think the relevant line is this: https://github.com/wch/r-source/blob/4b24a12b68394e8772f50909d69fb428e9ec0844/src/main/identical.c#L91, but the entry point for R's `identical` is [here](https://github.com/wch/r-source/blob/4b24a12b68394e8772f50909d69fb428e9ec0844/src/main/identical.c#L42) and that code does a lot more than what's necessary.

That looks easy enough indeed, also to test.

> Either way, a C++ function that does the lookup with a set is probably best.

Something I'll add then.

vandenman:
Okay, so there is now a separate branch for this. We can continue the discussion here or in `jasp-stats/INTERNAL-jasp#808`, but I'll close this PR for now.

> I think comparing to the "value" is a good choice

Yes, but because of that I can't use any of R's default lookup functions (those only work for strings).

> I'm pretty sure it is going to be easy to go from a symbol to a value, maybe as easy as calling "parse" or calling some function.

One way is `fun <- eval(parse(text = string))` but more efficient is probably `mget(c("list", "<-", "rnorm"), ifnotfound = NA, inherits = TRUE)`, where the part between `c(...)` can be any character vector. That actually answers my own question about how to construct the whitelist...

Let me know when you have something working that I can test!

PR #17, opened 2021-02-18
adjusted editImage to use jaspObjects

vandenman:
reorganizes common and makes some changes to `editImage` required for https://github.com/jasp-stats/jasp-desktop/pull/4470


R/common.R
```diff
@@ -17,16 +17,40 @@
...
+    other   = jaspResultsCPP$getOtherObjectsForState()
+  )
+
+  jaspResultsCPP$relativePathKeep <- .saveState(newState)$relativePath
+
+  returnThis <- NULL
+  if (returnKeepList)
+    returnThis <- list(keep=jaspResultsCPP$getKeepList()) #To keep the old keep-code functional we return it like this
+
+  jaspResultsCPP$complete() #sends last results to desktop, changes status to complete and saves results to json in tempfiles
```
JorisGoosen, line 39: `finishJaspResults` is now being used in `editImage` right?
So you should not only work around the keeplist but you should only call `complete` for an actual run of an analysis. And `jaspResultsCPP$saveResults()` and `$finishWriting()`  instead.

JorisGoosen (changes requested):

R/common.R
vandenman: I think I did this now, but let me know if this was not what you had in mind!

R/deleteThis.R
```diff
@@ -0,0 +1,1789 @@
+# these functions should be deleted. I couldn't find any usage in the modules or jasp-desktop
+
+jaspResultsStrings <- function() {
```
TimKDJ, line 3: This one is an "active" hack to get stuff from jaspResults translated. In JASP, jaspResults is not a package so it cannot use its own translation files

R/deleteThis.R
```diff
@@ -0,0 +1,1789 @@
...
+    seq <- c()
+  else
+    seq <- from:to
+
+  seq
+}
+
+#.clean is not necessary for analyses using jaspResults, jaspTable will take care of it for you.
+
+.parseMessage <- function(message, class, ...) {
```
TimKDJ, line 1251: This is used in `.hasErrors()` I think

R/deleteThis.R
```diff
@@ -0,0 +1,1789 @@
...
+  #
+  terms = sapply(modelTerms, function(x) paste0(unlist(x), collapse = ":"))
+  terms = terms[terms != ""]
+  formula.rhs = paste(terms, collapse = " + ")
+  if (formula.rhs != "") as.formula(paste(" ~ ", formula.rhs), env = env)
+}
+
+
+b64 <- function(x, ...) UseMethod("b64")   ## Translate names in x to 'base64'
+d64 <- function(x, ...) UseMethod("d64")   ## Untranslate names in x from 'base64'
```
TimKDJ, line 1534: Could you double check if these aren't used in the classical meta analysis?

R/deleteThis.R
```diff
@@ -0,0 +1,1789 @@
...
+  if (identical(Sys.getenv("TRAVIS"), "true")) {
+    return(suppressWarnings(suppressMessages(expr)))
+  } else {
+    return(expr)
+  }
+}
+
+# Operator to perform an expression by group / conditionally on a factor / given a condition. (author: EJvK):
+# Usage: (abs(mtcars$mpg - mean(mtcars$mpg)) > var(mtcars$mpg)) %|% mtcars$cyl
+`%|%` <- function(expr, group)
```
TimKDJ, line 1772: This one appears to be used by the filter/compute column interface

vandenman: I've added it to `friendlyConstructorFunctions.R` which is a somewhat better home for it than `common.R`.

TimKDJ: Yup sounds sensible

R/deleteThis.R
vandenman: I've moved the ones that are used back to common. But since the classical meta-analysis is the only analysis that uses these, it might be better to move these to that module.

R/deleteThis.R
vandenman: I've moved it to `commonmessages.R`.

R/deleteThis.R
vandenman: I've moved it back.

R/deleteThis.R
TimKDJ: Theoretically I could see the use of these in jaspBase, but I highly doubt anyone will ever use them

TimKDJ:
I'm confident enough that the functions in `deleteThis.R` are not being used that you could just delete them.

R/deleteThis.R
```diff
@@ -0,0 +1,1697 @@
...
+
+.seqx <- function(from, to) {
+
+  if (from > to)
+    seq <- c()
+  else
+    seq <- from:to
+
+  seq
+}
```
TimKDJ, line 1241: These are used in the ANOVA's

R/deleteThis.R
```diff
@@ -0,0 +1,1697 @@
...
+  ret
+}
+
+.cat <- function(object) {
+
+  cat(toJSON(object))
+
+}
+
+.dataFrameToRowList <- function(df, discard.column.names=FALSE) {
```
TimKDJ, line 1199: This one is found in the contingency tables

R/deleteThis.R
vandenman: I'll delete them there (especially considering the first one is just re-implementing `seq_along`)

R/deleteThis.R
vandenman: good find! I've copied it over to that module. Shall I also do that with the functions specific to the classical meta-analysis?

vandenman:
> I'm confident enough that the functions in deleteThis.R are not being used that you could just delete them.

I'd be happy to! But I figured this way it's a bit easier to check if I didn't make any mistakes (and apparently I made quite a few).

R/deleteThis.R
TimKDJ: Yeah, I think functions that have been around for years but are only used in a single module don't need to be in jaspBase. Be sure to update the NAMESPACE too

vandenman: done!

NAMESPACE
```diff
@@ -18,12 +18,9 @@ export(.diff)
 export(.unvf)
 export(.messages)
 export(.quietDuringUnitTest)
-export(.dataFrameToRowList)
```
TimKDJ, line 21: I'm guessing we can also get rid of:
- `.newFootnotes`
- `.addFootnote`
- `.diff`
- `.quietDuringUnitTest`

TimKDJ: And either we should move the classical meta analysis functions (like you suggested, which is a good idea) or they should be added here

vandenman: done

TimKDJ:
No more comments from me; it's nice to have a good cleanup like this.
If @JorisGoosen approves the PR can be merged

vandenman:
- [ ] Actually, this PR will make `rewriteImages` fail because that still uses `.writeImage`...

JorisGoosen:
>     * [ ]  Actually, this PR will make `rewriteImages` fail because that still uses `.writeImage`...

Aah, so I shouldn't merge it then?

vandenman:
I'll fix it first

PR #20, opened 2021-03-01
changes for custom fonts

vandenman:
The commented stuff inside `registerFonts` shows what we'd need to do to register font files that we ship with JASP.

### DIFF
--- a/R/common.R
+++ b/R/common.R
@@ -139,7 +139,6 @@ registerFonts <- function() {
 
     print("registered FreeSansJASP & LatoJASP")
     jaspGraphs::setGraphOption("family", .resultsFont)
-
   } else {
     print("R packages 'ragg' and/ or 'systemfonts' are unavailable, falling back to R's default fonts.")
   }
--- a/R/common.R
+++ b/R/common.R
@@ -127,18 +127,19 @@ registerFonts <- function() {
 
   if (requireNamespace("ragg") && requireNamespace("systemfonts")) {
 
-    # # Not sure if we actually want to do this
-    # font_name0 <- "FreeSansJASP"
-    # font_name1 <- "LatoJASP"
-    # # TODO: this obviously only works on my machine
-    # font_file0 <- "~/github/jasp-desktop/Desktop/resources/fonts/FreeSans.ttf"
-    # font_file1 <- "~/github/jasp-desktop/Desktop/html/font/Lato-Regular.ttf"
+    # To register custom font files shipped with JASP we need the path to the font file.
+    # Next the font could be loaded like this:
     #
-    # systemfonts::register_font(font_name0, normalizePath(font_file0))
-    # systemfonts::register_font(font_name1, normalizePath(font_file1))
+    # fontName <- "FreeSansJASP"
+    # fontFile <- "~/github/jasp-desktop/Desktop/resources/fonts/FreeSans.ttf"
+    # systemfonts::register_font(fontName, normalizePath(fontFile))
+    # jaspGraphs::setGraphOption("family", fontName)
 
-    print("registered FreeSansJASP & LatoJASP")
-    jaspGraphs::setGraphOption("family", .resultsFont)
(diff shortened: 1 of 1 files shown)


R/common.R
```diff
@@ -80,6 +80,7 @@ runJaspResults <- function(name, title, dataKey, options, stateKey, functionCall
     dataset <- do.call(.readDataSetToEnd, cols)
   }
 
+  registerFonts()
```
TimKDJ, line 83: This might be good to perform when a new R session is started inside `initEnvironment()`

R/common.R
```diff
@@ -122,6 +123,27 @@ runJaspResults <- function(name, title, dataKey, options, stateKey, functionCall
...
+    # font_name1 <- "LatoJASP"
+    # # TODO: this obviously only works on my machine
+    # font_file0 <- "~/github/jasp-desktop/Desktop/resources/fonts/FreeSans.ttf"
+    # font_file1 <- "~/github/jasp-desktop/Desktop/html/font/Lato-Regular.ttf"
+    #
+    # systemfonts::register_font(font_name0, normalizePath(font_file0))
+    # systemfonts::register_font(font_name1, normalizePath(font_file1))
+
+    print("registered FreeSansJASP & LatoJASP")
+    jaspGraphs::setGraphOption("family", .resultsFont)
```
TimKDJ, line 141: Where does `.resultsFont` come from?

TimKDJ (changes requested):

R/common.R
TimKDJ: From the other PR, I see.

R/common.R
vandenman: That would ensure that it always exists, which is a good idea! But to update it, the current constructions with the `jasp_rcpp_... ` functions might still be necessary.

R/common.R
vandenman: yes. We could also pass this as an argument but there's not a very clear benefit and it would add yet another argument to `runJaspResults`.

R/common.R
```diff
@@ -122,6 +123,28 @@ runJaspResults <- function(name, title, dataKey, options, stateKey, functionCall
   }
 }
 
+registerFonts <- function() {
+
+  if (requireNamespace("ragg") && requireNamespace("systemfonts")) {
```
JorisGoosen, line 128: Shouldn't you just add `ragg` and `systemfont` to the imports of jaspBase?

R/common.R
```diff
@@ -122,6 +123,28 @@ runJaspResults <- function(name, title, dataKey, options, stateKey, functionCall
...
+    # To register custom font files shipped with JASP we need the path to the font file.
+    # Next the font could be loaded like this:
+    #
+    # fontName <- "FreeSansJASP"
+    # fontFile <- "~/github/jasp-desktop/Desktop/resources/fonts/FreeSans.ttf"
+    # systemfonts::register_font(fontName, normalizePath(fontFile))
+    # jaspGraphs::setGraphOption("family", fontName)
+
+    if (exists(".resultsFont"))
+      jaspGraphs::setGraphOption("family", .resultsFont)
```
JorisGoosen, line 139: And in fact `jaspGraphs`? 

For renv I have removed the default installation of jaspGraphs in the delivered R-library. Because renv  will install the particular version needed by each pkg. Would that break this code?

JorisGoosen: I guess it would actually make more sense to store the desired font in `jaspBase` and then have `jaspGraphs` query that. This way we make the dependency-graph a bit more linear.

JorisGoosen (changes requested):

R/common.R
JorisGoosen: > 
> 
> This might be good to perform when a new R session is started inside `initEnvironment()`

https://github.com/jasp-stats/INTERNAL-jasp/issues/1345

JorisGoosen (approved):
Well I think this could work.

PR #22, opened 2021-03-04
implemented a checkSum and ensure package versions in descriptions are followed

vandenman:
1. Made the installation of modules with Description only even slower but now ensures that the installed package versions obey the DESCRIPTION file.

2. Added an md5 checksum. This creates one md5 checksum for each file (according to a pattern).


JorisGoosen:
this was manually cherry-picked

PR #27, opened 2021-03-29
Renv

JorisGoosen:
Necessary for the renv PR in jasp-desktop to function.


R/moduleInstall.R
```diff
@@ -1,12 +1,338 @@
+postInstallFixes <- function(folderToFix) {
+  if(length(ls(all.names=TRUE,pattern="\\..postProcessLibraryModule")) > 0 ) #We seem to be running in JASP
+  {
+    #print("we are *in* jasp, so we use .postProcessLibraryModule!")
+    .postProcessLibraryModule(folderToFix)
+  }
+  else 
+  {
+    #We do not have that function available so we will need to start JASPEngine ourselves, but where is it?
+    jaspEngineLocation <- ifElse(Sys.getenv("JASPENGINE_LOCATION") != "", Sys.getenv("JASPENGINE_LOCATION"), paste0(getwd(), "/../JASPEngine"))
```
vandenman, line 10: more ideomatic would be
```suggestion
    jaspEngineLocation <- Sys.getenv("JASPENGINE_LOCATION", unset = file.path(getwd(), "..", "JASPEngine"))
```
without that, it probably also doesn't work on Windows, although we probably don't need this there anyway.

R/moduleInstall.R
```diff
@@ -1,12 +1,338 @@
...
+
+  # TODO: unclear whether this is necessary within JASP, or just within Rstudio.
+  # renv must be unaware of any other libPaths than the cache and the directory designated for the module.
+  # if there are other libPaths then renv may reuse pkgs from those other libPaths
+  # it does copy those pkg to the cache before symlinking them
+  # inspired by https://stackoverflow.com/a/36873741/4917834
+  # it does appear to be necessary within rstudio and when the pkgs from jasp-required-files are present in a libPath
+  old.lib.loc <- .libPaths()
+  on.exit(assign(".lib.loc", old.lib.loc,   envir = environment(.libPaths)))
+          assign(".lib.loc", moduleLibrary, envir = environment(.libPaths))
```
vandenman, line 79: We should probably test to what extent it is necessary to keep this in there. I added it originally so renv wouldn't copy packages from my local R library, in other words, force renv to use whatever we give it. But if that's already guaranteed by JASP through other ways then maybe we should delete this.

R/moduleInstall.R
```diff
@@ -1,12 +1,338 @@
...
+  print(sprintf("Installing module with renv. installJaspModuleFromRenv('%s', c(%s), '%s', '%s', %s)",
+                modulePkg, paste0("'", libPathsToUse, "'", collapse = ", "), moduleLibrary, repos, onlyModPkg))
+
+  renv::consent() #Consent to doing stuff in certain paths: https://rstudio.github.io/renv/reference/consent.html We've changed at least the root and the cache so hopefully that means it won't be changing stuff in the locations it mentions
+  options(renv.verbose=TRUE) # More feedback wanted although this seems to do little
+  options(renv.config.install.verbose=TRUE)
+
+  if (!dir.exists(moduleLibrary))
+    if (!dir.create(moduleLibrary))
+      stop("failed to create moduleLibrary!")
```
vandenman, line 57: Something I found out only recently is that all strings passed to `stop` and `warning` end up in weblate for translation... I assume that some of this info is only for logging purposes? In that case, it doesn't make sense to have a Chinese error message here. We can ensure that these messages won't be translated by adding `domain = NA` as an argument.

R/moduleInstall.R
```diff
@@ -1,12 +1,338 @@
...
+  if (!file.exists(pkgDescr))
+    stop("Your module contains no DESCRIPTION file")
+
+  return(read.dcf(file.path(modulePkg, "DESCRIPTION"))[1, ])
+
+}
+
+renv_diagnostics_packages_as_df <- function(project) {
+
+  lockfile  <- renv:::renv_diagnostics_packages_lockfile(project)
```
vandenman, line 219: perhaps we should add a comment that this function is equivalent to `renv:::renv_diagnostics_packages` except that it returns a data.frame instead of directly printing all information (and returning NULL).

R/moduleInstall.R
```diff
@@ -1,12 +1,338 @@
...
+  options(install.opts = "--no-docs --no-test-load"); #make sure we do not do a test-load, because this will not work on mac. the rpaths still need to be fixed
+
+  #Try to nudge renv towards installing binaries when possible
+   if(getOS() == "windows" || getOS() == "osx")
+    options(pkgType = "binary");
+  
+  addRenvBeforeAfterDispatch()
+}
+
+installJaspModuleFromDescriptionOld <- function(modulePkg, libPathsToUse, moduleLibrary, repos, onlyModPkg) {
```
vandenman, line 328: is this function still used? If not, we can delete it?

vandenman (changes requested):
some minor things, see below.

R/checkSums.R
```diff
@@ -0,0 +1,37 @@
...
+    list.files(file.path(modulePkg, "R"),         recursive=TRUE, full.names = TRUE, pattern = "\\.R$"),
+	list.files(file.path(modulePkg, "inst"),      recursive=TRUE, full.names = TRUE, pattern = "\\.(qml|po|svg|png|jpg|md)$"),
+    list.files(modulePkg,                         recursive=TRUE, full.names = TRUE, pattern = "renv.lock")
+  )
+  newMd5Sums <- tools::md5sum(srcFiles)
+  newMd5Sums
+}
+
+makeMd5SumsFilename <- function(modulePkg, moduleLibrary) {
+  file.path(paste0(moduleLibrary, "/.."), paste(basename(modulePkg), "md5sums.rds", sep = "_"))
```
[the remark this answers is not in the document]
JorisGoosen: No actually that will be fine, R fixes all the paths al the time anyway.

R/moduleInstall.R
JorisGoosen: yeah I think there is a whole bunch more stuff that can be scrapped. Also the whole required-modules stuff can be gone. Because renv fixes all of that now.

But I first wanted to get a functional version up. After that I will start cleaning up the dynamic module code and then this can possibly be taken with. Or separately for that matter.

R/moduleInstall.R
JorisGoosen: Aah that is annoying yeah. This stuff is just for logging indeed.

JorisGoosen:
@vandenman your suggestions sound good, Ive applied one through github. For the other ones. you could just edit it and push to `upstream/renv`, this will probably be more efficient than me reading here what you read here and applying what your write here :p

R/moduleInstall.R
```diff
@@ -1,12 +1,338 @@
+postInstallFixes <- function(folderToFix) {
+  if(length(ls(all.names=TRUE,pattern="\\..postProcessLibraryModule")) > 0 ) #We seem to be running in JASP
+  {
+    #print("we are *in* jasp, so we use .postProcessLibraryModule!")
+    .postProcessLibraryModule(folderToFix)
+  }
+  else
```
TimKDJ, line 7: Is the `else` for use outside of JASP? Any reason in particular we use system variables rather than an optional argument to `postInstallFixes`?

TimKDJ (commented):
The code looks good to me. Most of my comments are more about increasing my understanding to be honest. I suppose I had better read up on how `renv` works exactly!

R/moduleInstall.R
```diff
@@ -1,12 +1,338 @@
...
+    dir.create(file.path(moduleLibraryTemp, "renv"), recursive = TRUE)
+
+  setupRenv(moduleLibrary)
+
+  # make renv blind for other libPaths
+  old.lib.loc <- .libPaths()
+  on.exit(assign(".lib.loc", old.lib.loc, envir = environment(.libPaths)))
+  assign(".lib.loc", moduleLibrary, envir = environment(.libPaths))
+
+  # TODO: this is not very efficient because renv::install looks up the remotes on github...
```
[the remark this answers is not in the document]
vandenman: Not if you add a GitHub token. I don't think there is any way around this.

R/moduleInstall.R
```diff
@@ -1,12 +1,338 @@
...
+    renv::restore(
+      project  = moduleLibraryTemp,
+      library  = moduleLibrary,
+      lockfile = lockFileTemp,
+      clean    = TRUE,
+      prompt   = prompt
+    )
+
+    if (!libraryMatchesLockfile(moduleLibraryTemp))
+      warning("Failed to recreate lock file of module!")
```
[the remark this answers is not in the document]
vandenman: It means that renv had to improvise and could not exactly replicate a lockfile as is. This can happen if you use package versions that are not available anywhere, or cannot install old packages from source without any available binary. In that scenario, renv will just try to make things work with other package versions (which I think is reasonable).

I don't think this will happen ever if:
- a module doesn't use a lockfile but only a DESCRIPTION
- a module updates its dependencies somewhat regularly (say twice a year).

R/moduleInstall.R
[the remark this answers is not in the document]
JorisGoosen: You definitely want to set up a GITHUB_PAT otherwise lots and lots of tears. Sadly enough I experienced this a lot :(
Too many computers and PATs.

R/moduleInstall.R
```diff
@@ -1,12 +1,338 @@
...
+
+  for(cachePath in cachePaths)
+    if (!dir.exists(cachePath))
+     stop(sprintf("A cache is supposed to be at '%s' but it does not exist!", cachePath))
+
+  Sys.setenv("RENV_PATHS_LIBRARY" = moduleLibrary)
+
+  print("Using the following paths:")
+  for(name in names(renv::paths))
+    print(sprintf("%s:%s%s", name, paste0(rep(" ", 12 - nchar(name)), collapse=""), renv::paths[[name]]()))
```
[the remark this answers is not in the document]
JorisGoosen: Hmm, no this looks like etermely suspect code... @vandenman ?

R/moduleInstall.R
```diff
@@ -1,12 +1,338 @@
...
+    {
+      installPath <- mget("installpath", envir = parent.frame(1), ifnotfound = "unknown")
+
+      if(installPath != "unknown") 
+      {
+        print(sprintf("Installed %s to '%s', now running post install fixes.", x, installPath))
+        postInstallFixes(installPath)
+      } 
+      else
+        print(sprintf("Installing %s failed for some reason...", x))
```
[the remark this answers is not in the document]
JorisGoosen: I think I've seen this message pop up a lot before `renv` decides to start looking at remotes, so it might be that the error msg is just a bit misguided.

R/moduleInstall.R
JorisGoosen: Yeah when you are building JASP ^^

TimKDJ: Makes sense! ^.^

JorisGoosen:
Im going to use git to merge this manually

PR #28, opened 2021-05-03
Set right Working Directory when calling loadJaspResults

boutinb:
Probably solves https://github.com/jasp-stats/INTERNAL-jasp/issues/1365

### DIFF
--- a/R/common.R
+++ b/R/common.R
@@ -18,6 +18,12 @@ fromJSON <- function(x) jsonlite::fromJSON(x, TRUE, FALSE, FALSE)
 toJSON   <- function(x) jsonlite::toJSON(x, auto_unbox = TRUE, digits = NA, null="null")
 
 loadJaspResults <- function(name) {
+  if (base::exists(".requestStateFileNameNative")) {
+	location              <- .fromRCPP(".requestStateFileNameNative")
+	oldwd                 <- getwd()
+	setwd(location$root)
+	on.exit(setwd(oldwd))
+  }
   jaspResultsModule$create_cpp_jaspResults(name, .retrieveState())
 }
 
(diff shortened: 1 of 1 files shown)


R/common.R
```diff
@@ -18,6 +18,12 @@ fromJSON <- function(x) jsonlite::fromJSON(x, TRUE, FALSE, FALSE)
 toJSON   <- function(x) jsonlite::toJSON(x, auto_unbox = TRUE, digits = NA, null="null")
 
 loadJaspResults <- function(name) {
+  if (base::exists(".requestStateFileNameNative")) {
+	location              <- .fromRCPP(".requestStateFileNameNative")
+	oldwd                 <- getwd()
+	setwd(location$root)
+	on.exit(setwd(oldwd))
```
vandenman, line 25:
```suggestion
  if (exists(".requestStateFileNameNative")) {
    location              <- .fromRCPP(".requestStateFileNameNative")
    oldwd                 <- getwd()
    setwd(location$root)
    on.exit(setwd(oldwd))
```
fixed formatting, we prefer 2 spaces for indentation in R.

vandenman (changes requested):
1. Does this actually work/ did you test this?
2. Can the other call to `location <- .fromRCPP(".requestStateFileNameNative")` be deleted?
3. Why do we actually do the part with `on.exit()?` what goes wrong if we don't restore the working directory?

The main issue with this code is that `on.exit` runs once the current function finishes. That means that the `setwd` doesn't persist but only happens for that function call and is reset afterward. 

The easy way around that is if there is actually no need to do `on.exit` to restore the working directory.

There is also a difficult way but if it's not necessary then I'd rather not do that.

PR #31, opened 2021-05-21
Add the jaspResults "gridGraphics" dependency

TimKDJ:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/1268

@vandenman: it might be better to put the jaspResults dependencies here considering jaspBase is a common dependency of all modules

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -8,11 +8,11 @@ Description: Package contains the JASP Bayesian and Frequentist analyses.
     See <https://jasp-stats.org> for more details.
 License: GPL2+
 Imports: 
-	modules,
-	pkgbuild,
-	renv,
-	remotes, 
-	jsonlite,
-	rjson,
-	Rcpp
-
+  gridGraphics,
+  modules,
+  pkgbuild,
+  renv,
+  remotes, 
+  jsonlite,
+  rjson,
+  Rcpp


TimKDJ:
Ideally jaspResults dependencies would just go into the jaspResults DESCRIPTION though. Could we do a `remotes::install_deps("/jaspResults")` or something into the global renv package cache?

vandenman:
So this is fine for developing in R but it won't actually fix jasp-stats/jasp-test-release#1268 because jaspResults is not an R package inside JASP and no module depend on it as a package. I think that a better band-aid for now is to add the dependency to jaspSummaryStats (in fact, multiple modules have done that already).

> Could we do a remotes::install_deps("/jaspResults") or something into the global renv package cache?

You can't install stuff into the global cache directly, but you probably could probably read the DESCRIPTION of jaspResults and install the packages that way.

TimKDJ:
This is jaspBase, though, not jaspResults.

vandenman:
> This is jaspBase, though, not jaspResults.


Oops

vandenman (approved):

TimKDJ:
> You can't install stuff into the global cache directly, but you probably could probably read the DESCRIPTION of jaspResults and install the packages that way.

I'll make an issue for it

vandenman:
well if the dependency is in jaspBase then it will get installed automagically. Ideally, jaspResults just becomes a proper R package at one point and then this is all unnecessary.

TimKDJ:
> well if the dependency is in jaspBase then it will get installed automagically.

Yup, that's why I added it here. Feels a bit hacky though.

> Ideally, jaspResults just becomes a proper R package at one point and then this is all unnecessary.

Definitely. Translations are also an issue when jaspResults is not a package.

PR #33, opened 2021-05-26
Add missing jaspBase dependencies and also add all jaspResults dependencies

TimKDJ:
So from what I can tell they use:

**jaspResults:**
- Rcpp,
- R6,
- BH,
- ggplot2,
- gridGraphics,
- qgraph,
- ragg

**jaspBase:**
- gbm,
- ggplot2,
- gridExtra,
- jsonlite,
- officer,
- ragg,
- rjson,
- rvg,
- stringr,
- systemfonts,
- plyr,
- codetools,
- desc,
- pkgbuild,
- remotes,
- renv,
- withr

I've not included jaspGraphs here as we have workarounds for that in JASP and jaspTools.
I'm also not entirely sure whether to include:
- gbm? is only used in ml
- ragg & systemfonts? have a workaround if they don't exist, but shouldn't we just have them
- desc? are the `getDepsFromDescription` fns used?

Can you have a look at the list @vandenman?


TimKDJ:
Not sure if jaspResults' dependencies should go here or if we should implement https://github.com/jasp-stats/INTERNAL-jasp/issues/1381

vandenman:
So using `renv::dependencies` I get these lists (bold is new). I sometimes added a few comments.

**jaspResults:**
- BH
- ggplot2
- **grDevices**
- **grid**
- gridGraphics
- **methods** 
- qgraph
- R6
- ragg
- Rcpp

**jaspBase:**
- codetools
- **compiler** (`compiler::enableJIT` on windows, maybe we can remove this?)
- desc
- gbm
- ggplot2
- grDevices
- gridExtra
- jaspTools (used in `.saveState`, we can probably ignore this)
- jsonlite
- **modules** (only used in the DESCRIPTION but I think that @JorisGoosen uses this inside JASP. I'm not sure if we still need this now that everything is a package)
- officer
- pkgbuild
- plyr
- ragg
- Rcpp
- remotes (only used in `installJaspModuleFromDescriptionOld`, which we can delete I guess).
- renv
- rjson (this should be replaced by `fromJSON` and `toJSON` that we use everywhere else. This is also only used in `runJaspResults` for the dataKey, options, and stateKey)
- rvg 
- stats
- stringr
- systemfonts
- tools
- withr

vandenman:
I think we should be that strict about package dependencies.

vandenman:
oh and `desc` can also go I think, as well as the functions `getDepsFromDescription` and `.parseDescription`, I don't think these are used anywhere.

TimKDJ:
I removed `desc`, `remotes`, `gbm` and `stringr` and added the base R packages.
I left both json packages, because I know that they show different behaviour in certain situations (`jaspTools` also uses both) and I don't really want to dive into that right now.

vandenman (approved):
Changes look fine! 

Maybe @JorisGoosen should also take a quick look at whether this PR won't break the build bot (again)?

Concerning:

>  I left both json packages, because I know that they show different behaviour in certain situations (jaspTools also uses both) and I don't really want to dive into that right now.

Sure, that's fine for now, it's probably a lot more work to change (and test) that properly. In the long run I would prefer to use `jsonlite` in favor of `rjson` everywhere, because the former seems to be better maintained (more frequently updated + more credible maintainer).

JorisGoosen:
Well, you don really need me for that, you can just check [the macos req-files](https://github.com/jasp-stats/jasp-required-files/tree/MacOS/Frameworks/R.framework/Versions/4.0/Resources/library) and [the windows one](https://github.com/jasp-stats/jasp-required-files/tree/Windows/R/library) and see if the added packages are there.

I see tthat `BH` for instance is missing, so Im guessing there are more.

If this is merged without adding them the  buildbot won't be functional. And it might be a while before I notice cause Im in a forest and haven't set up a connection up back home yet.

PR #37, opened 2021-07-13
Add editOptions in the response of the editImage function

boutinb:


R/common.R
```diff
@@ -899,7 +899,8 @@ editImage <- function(name, optionsJson) {
       width    = width,
       height   = height,
       revision = revision,
-      error    = FALSE
+      error    = FALSE,
+      editOptions = jaspGraphs::plotEditingOptions(plot)
```
vandenman, line 903: So performance isn't really an issue here because `jaspGraphs::plotEditingOptions` will return a cached set of options if that exists, rather than recomputing them. Nevertheless, this solution is doing things twice.

After `jaspPlotCPP$plotObject <- newPlot`, `writeImage` always creates `editOptions`. Is it possible to reuse that? The underlying issue here is probably that we're doing `jaspPlotCPP$convertToJSON()`, manually, in R. Maybe there is a better way? Because I guess things went wrong now because `jaspPlotCPP$convertToJSON()` was updated but the fields in `response` were not.

boutinb: The goal is here to have the uptodate editOptions, especially when resetPlot is used: in this case only the jaspGraphs knows what are the default options.

vandenman (approved):

PR #42, opened 2021-07-28
Reenable save as svg and improve error message

vandenman:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/1390

The changes to `saveImage` are identical to those in https://github.com/jasp-stats/jasp-desktop/pull/4215, but it got lost when we split off jaspBase into a module. For this to work, we also need `svglite` as a dependency and that's why I assigned @JorisGoosen as a reviewer, who can hopefully tell me what I need to do to ensure that the buildbot/ nightlies don't break.

In addition, the error message about invalid format now says what format is actually invalid, and `domain = NA` avoids this from being ever translated.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -28,5 +28,6 @@ Imports:
   renv,
   rjson,
   rvg,
+  svglite,
   systemfonts,
   withr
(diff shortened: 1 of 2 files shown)


JorisGoosen:
I'll add `svglite` in `jasp-required-files` for both windows and macos.

The flatpak script will handle adding it to the helper zip right?

JorisGoosen (approved):

vandenman:
yes, the flatpak script will do that automagically

PR #54, opened 2021-11-16
Fixed a typo! 🖊

amirmasoudabdol:
Unless we went for French there! 😬

### DIFF
--- a/R/moduleInstall.R
+++ b/R/moduleInstall.R
@@ -25,7 +25,7 @@ installJaspModule <- function(modulePkg, libPathsToUse, moduleLibrary, repos, on
     moduleName <- getModuleInfo(modulePkg)[["Package"]]
     if (dir.exists(file.path(moduleLibrary, moduleName))) {
       print(sprintf("Nothing changed according to md5sums, not reinstalling %s.", moduleName))
-      return("succes!")
+      return("Success!")
     } else {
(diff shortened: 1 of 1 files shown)


JorisGoosen:
Or dutch for that matter ;)

JorisGoosen:
But doesnt this break 
```
	std::string		result			= jaspRCPP_evalRCode(moduleCode.c_str(), false);
	bool			succes			= result == "succes!"; //Defined in DynamicModule::succesResultString()
```
at https://github.com/jasp-stats/jasp-desktop/blob/ef60937d81e5388ed2f27f6047cc1363e9af4611/Engine/engine.cpp#L442 ?
Im not sure about it but it is quite likely.

Im not happy about strings to communicate succes etc, but R is not easily coupled with `class enum` etc.

amirmasoudabdol:
That looks like that it will break! If the success and failure are the only two messages, this might makes sense to just use 0, and 1, and they can be casted to bool easily anyway?

JorisGoosen:
> That looks like that it will break! If the success and failure are the only two messages, this might makes sense to just use 0, and 1, and they can be casted to `bool` easily anyway?

possibly but it is a lot easier to do so later because if this is merged we break `development`.
And I don't mind a bit of dutch/french in the codebase. Much better than this tongue.

amirmasoudabdol:
Sure, it's not a big deal anyway. I just noticed it and made a quick PR on GitHub!

PR #56, opened 2021-12-06
Properly export S3 methods so filter + computed columns can find them

vandenman:
Fixes https://github.com/jasp-stats/jasp-issues/issues/1507

The problem is that we did not export `ifElse`, `replaceNA`, and `zScores` as S3 methods. So you'd have to call them as individual functions. For example, rather than having `zScores(...)` dispatch to the appropriate method you'd have to do `zScores.numeric(...)` which completely defeats the purpose of S3 dispatch.

In my opinion, the bigger issue is that we do not use roxygen2 to generate the NAMESPACE. For example, the current setup pollutes the export by exporting `Ifelse.integer`, `Ifelse.numeric`, etc., because at the top of the NAMESPACE we do `exportPattern("^[[:alpha:]]+")`.

@TimKDJ If you prefer, I'd be more than happy to rewrite the current situation with roxygen2. This is a must anyway if we want to generate online documentation in the same way as I did for jaspGraphs.


TimKDJ:
Yeah I think doing this properly with roxygen would be nicer.
It'd be good to do that for all R modules at some point, but let's start small :)

vandenman:
Okay, I'll redo this with roxygen2 and will let you know once I'm finished.

vandenman:
@TimKDJ should be good for review!

vandenman:
I forgot to export `VovkSellkeMPR`, but I guess it's also nice to see that the action works as intended :stuck_out_tongue:

TimKDJ:
> I forgot to export `VovkSellkeMPR`, but I guess it's also nice to see that the action works as intended 😛

Yeah nice to see it in action and actually catching an error!

TimKDJ (changes requested):
I don't think the following exports are needed (not entirely sure about the image ones):

```
findDominatingClass
reorderFactor
saveImage
rewriteImages
editImage
fakeGrDevicesPdf
```

And `.vdf` does require an export.

vandenman:
```r
findDominatingClass
reorderFactor
fakeGrDevicesPdf
```
Good catch! I've removed these and added `.vdf`.


```r
saveImage
rewriteImages
editImage
```
These are exported because inside JASP we do things like
```c++
	staticResult = jaspRCPP_parseEvalStringReturn("saveImage(plotName, format, height, width)", true);
```
See https://github.com/jasp-stats/jasp-desktop/blob/20585b1da8c609f7f23be3206fb2e13f4f6bba1c/R-Interface/jasprcpp.cpp#L368, https://github.com/jasp-stats/jasp-desktop/blob/20585b1da8c609f7f23be3206fb2e13f4f6bba1c/R-Interface/jasprcpp.cpp#L385, and https://github.com/jasp-stats/jasp-desktop/blob/20585b1da8c609f7f23be3206fb2e13f4f6bba1c/R-Interface/jasprcpp.cpp#L403.

Now, I agree that there *should* be no need to export these. We could just do 
``` 
	staticResult = jaspRCPP_parseEvalStringReturn("jaspBase:::saveImage(plotName, format, height, width)", true);
```
Any preferences?

TimKDJ:
> Now, I agree that there should be no need to export these. We could just do

Yeah that'd be better IMO -- exposing this stuff in R is not great.

In the long run I could see us split off `jaspBase` in two packages, one that is concerned with installing/running inside JASP (e.g., `jaspEngine`) and one that contains all the building blocks for an analysis (`jaspBase`). The latter should also really contain the functions of `jaspResults`. Someday, hopefully.

vandenman:
@TimKDJ Ok so we no longer export `saveImage`, `rewriteImages`, and `editImage`. jasp-stats/jasp-desktop#4691 should be merged before this PR, otherwise, a lot of stuff will break.

TimKDJ (approved):

PR #61, opened 2022-03-18
add power transformation functions

Kucharssim:
first step to implement: https://github.com/jasp-stats/INTERNAL-jasp/issues/1728

(the second step is adding the new functions to the R whitelist and computecolumnswindow in jasp-desktop)

A question: Should we perhaps set up tests for these functions in jaspBase as well? Perhaps I should also add more detailed documentation, assuming we will be exporting these functions anyway for the syntax mode?


vandenman (approved):
looks good to me!

JorisGoosen:
This can be merged right?

At least, I fail to see any way in which this will break either `stable` or `development` as it seems quite self-contained.

PR #63, opened 2022-05-03
deprecate .v() and .unv() explicitly

Kucharssim:
Despite we deprecated this a while ago and removed it from the majority of our code, I still see new JASP programmers attempt to use those functions sometimes. So I figured it could be nice to be explicit about this, instead of having to repeatedly ask to remove this during code review (we could do this more systematically in general whenever we deprecate a function/argument). Since `lifecycle::deprecate_warn()` generates the warning only once every 8 hours, this will not clog the output of the unit tests for analyses that still use these functions.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -16,6 +16,7 @@ Imports:
   gridExtra,
   gridGraphics,
   jsonlite,
+  lifecycle,
   methods,
   modules,
   officer,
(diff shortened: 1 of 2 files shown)


vandenman (approved):
Looks good to me! This is a pretty safe PR, but let's not jinx things (the installation of jaspBase is a bit fragile at the moment) and postpone this until 0.16.4.

PR #65, opened 2022-06-22
add logit and inverse logit

Kucharssim:
part of https://github.com/jasp-stats/INTERNAL-jasp/issues/1866 (after merging this, I still need to add this to the compute columns in jasp-desktop)


R/transformFunctions.R
```diff
@@ -72,3 +72,9 @@ YeoJohnson <- function(x, lambda) {
 
   return(result)
 }
+
+#' @export
+logit <- function(x) { stats::qlogis(x) }
+
+#' @export
+invLogit <- function(x) { stats::plogis(x) }
```
vandenman, line 80: how about
```suggestion
#' @export
logit <- stats::qlogis

#' @export
invLogit <- stats::plogis
```
?

Kucharssim: that would work fine of course, but it would also expose additional arguments of the `*logis` functions, which may be confusing when we make these functions available in the syntax mode

vandenman (approved):

PR #69, opened 2022-08-02
Add correct language recognition for repositories

shun2wang:
should fix https://github.com/jasp-stats/INTERNAL-jasp/issues/1898

before
image

after this PR
image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*.po linguist-detectable=true
+*.1 linguist-detectable=false


vandenman:
So this looks pretty good to me, but I wonder if we should add this to https://github.com/jasp-stats/.github instead so that it is added by default to all repositories.  Alternatively, we bite the bullet, add it manually to all existing repositories, and add it to https://github.com/jasp-stats/jaspModuleTemplate/ so that new modules have it by default.

@JorisGoosen any preferences?

shun2wang:
> but I wonder if we should add this to https://github.com/jasp-stats/.github 

but i can't confirm that .gitattributes file will work well in /.github path, because it may have some precedence such as $GIT_DIR/info/ and root path?
I don't know if my understanding is wrong，[see](https://git-scm.com/docs/gitattributes)

>when deciding what attributes are assigned to a path, Git consults $GIT_DIR/info/attributes file (which has the highest precedence), .gitattributes file in the same directory as the path in question, and its parent directories up to the toplevel of the work tree (the further the directory that contains .gitattributes is from the path in question, the lower its precedence). Finally global and system-wide files are considered (they have the lowest precedence).

JorisGoosen:
> So this looks pretty good to me, but I wonder if we should add this to [jasp-stats/.github](https://github.com/jasp-stats/.github) instead so that it is added by default to all repositories. Alternatively, we bite the bullet, add it manually to all existing repositories, and add it to [jasp-stats/jaspModuleTemplate](https://github.com/jasp-stats/jaspModuleTemplate/) so that new modules have it by default.
> 
> @JorisGoosen any preferences?

Well if just setting it in [jasp-stats/.github](https://github.com/jasp-stats/.github) sets it everywhere then that would be the best, if we need to change it it is a single point. Much preferable to setting it everywhere.

Of course if that doesnt work we have no choice and we can add it to everything. Which should be easy with a recursive call to `git submodules ...`

JorisGoosen:
We can just try it and see ofc

vandenman:
@shun2wang do you want to add this to https://github.com/jasp-stats/.github or shall I?

shun2wang:
you can do this please because i can't get a successful attempt at the /.github folder，I think you have write access for jaspBase so you can try it more flexible. and i will colse this PR for now.

with thanks

PR #71, opened 2022-08-04
allow for passing frameworklibrary to renv

JorisGoosen:
this also sets `sources` in renv::hydrate thus avoiding the use of systemwide or sitespecific libraries in R

Which caused quite some havoc on macos

### DIFF
--- a/R/moduleInstall.R
+++ b/R/moduleInstall.R
@@ -25,7 +25,7 @@ postInstallFixes <- function(folderToFix) {
 }
 
 #' @export
-installJaspModule <- function(modulePkg, libPathsToUse, moduleLibrary, repos, onlyModPkg, force = FALSE, cacheAble=TRUE) {
+installJaspModule <- function(modulePkg, libPathsToUse, moduleLibrary, repos, onlyModPkg, force = FALSE, cacheAble=TRUE, frameworkLibrary=NULL) {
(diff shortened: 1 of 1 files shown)


R/moduleInstall.R
```diff
@@ -156,8 +156,11 @@ installJaspModuleFromDescription <- function(modulePkg, libPathsToUse, moduleLib
 
   # TODO: this is not very efficient because renv::install looks up the remotes on github...
   # there is a better way but it requires us to mess with renv's internals or to be more explicit about pkgs
-  renv::hydrate(library = moduleLibrary, project = modulePkg)
-  renv::install(project = modulePkg, library = moduleLibrary, prompt = prompt)
+  # R gurus: the following if-else can be prettier no doubt...
+  if(!is.null(frameworkLibrary))  { sources <- c(moduleLibrary, frameworkLibrary)
+  } else                          { sources <- c(moduleLibrary)                   }
```
Kucharssim, line 161: so I am not actually sure whether you need this if-else at all, as

```r
c("something", NULL)
```

returns

```r
> "something"
```

JorisGoosen: that would def make it much cleaner

Kucharssim (approved):

PR #75, opened 2022-09-06
Restore options after running an analysis

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1833

@JorisGoosen about [this comment](https://github.com/jasp-stats/INTERNAL-jasp/issues/1833#issuecomment-1212037257)

> This could be combined with `jaspBase:::.cleanEngineMemory()` right?

I'd propose the opposite :smile:. `.cleanEngineMemory()` can be called from `setOptionsCleanupHook`. That way, we'll also ensure in the future that people running syntax won't get their global environment polluted by some weird jasp/ R package function. Also, the less C++ <-> R back and forth interaction, the better IMO.

While the changes here are probably quite harmless, I nevertheless think it's best to merge this after 0.16.4 since this only cleans up some code and doesn't fix anything at the moment (hence I assigned it to 0.17). After this PR is merged, https://github.com/jasp-stats/jaspAnova/pull/156 should probably be reverted.

### DIFF
--- a/R/common.R
+++ b/R/common.R
@@ -101,8 +101,9 @@ runJaspResults <- function(name, title, dataKey, options, stateKey, functionCall
   }
 
   registerFonts()
-  oldGraphOptions <- jaspGraphs::graphOptions()
-  on.exit(jaspGraphs::graphOptions(oldGraphOptions), add = TRUE)
+
+  # resets jaspGraphs::graphOptions & options after this function finishes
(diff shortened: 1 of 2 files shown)


Kucharssim (approved):
This looks fine to me, let's merge it after the release of 0.16.4 then!

Kucharssim:
Do you want to merge this now, @vandenman?

PR #76, opened 2022-09-15
extract whole error message

Kucharssim:
fixes https://github.com/jasp-stats/jasp-test-release/issues/1960

The previous method attempts to extract the error message by converting the error object into a character and parsing the string. This works fine as long as the error message itself does not contain symbol `:`, e.g.

```r
foo <- function() { stop("No components with an eigenvalue > 1.2. Maximum observed eigenvalue: 1.059") }
x <- try(foo())
```

Here we just extract the message directly from the object's attributes without relying on character parsing.

old implementation:

```r
.extractErrorMessage(x)
# [1] "1.059"
```

new implementation:

```r
.extractErrorMessage(x)
# [1] "No components with an eigenvalue > 1.2. Maximum observed eigenvalue: 1.059"
```

### DIFF
--- a/R/common.R
+++ b/R/common.R
@@ -536,10 +536,8 @@ jaspResultsStrings <- function() {
 
 #' @export
 .extractErrorMessage <- function(error) {
-
-  split <- base::strsplit(as.character(error), ":")[[1]]
-  last <- split[[length(split)]]
-  trimws(last)
+  msg <- attr(error, "condition")$message
+  trimws(msg)
 }
 
 #' @export
--- a/R/common.R
+++ b/R/common.R
@@ -546,7 +546,7 @@ jaspResultsStrings <- function() {
     last <- split[[length(split)]]
     return(trimws(last))
   } else {
-    stop("Do not know what to do with an object of class `", class(error)[1], "`; The class of the `error` object should be `try-error` or `character`!" )
+    stop("Do not know what to do with an object of class `", class(error)[1], "`; The class of the `error` object should be `try-error` or `character`!", domain = NA)
   }
 }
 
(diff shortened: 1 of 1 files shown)


vandenman:
this works, but are you sure that nobody puts a string directly into `.extractErrorMessage`? :sweat_smile:

Kucharssim:
> this works, but are you sure that nobody puts a string directly into .extractErrorMessage? 😅

No I am not sure as that did not occur to me (not sure what would be the use case). I could do a simple if switch for cases like this if you want :)

R/common.R
```diff
@@ -536,10 +536,18 @@ jaspResultsStrings <- function() {
...
-  trimws(last)
+  if (isTryError(error)) {
+    msg <- attr(error, "condition")$message
+    return(trimws(msg))
+  } else if (is.character(error)){
+    split <- base::strsplit(error, ":")[[1]]
+    last <- split[[length(split)]]
+    return(trimws(last))
+  } else {
+    stop("Do not know what to do with an object of class `", class(error)[1], "`; The class of the `error` object should be `try-error` or `character`!" )
```
vandenman, line 549:
```suggestion
    stop("Do not know what to do with an object of class `", class(error)[1], "`; The class of the `error` object should be `try-error` or `character`!", domain = NA)
```
otherwise translators will be asked to translate this message :)

vandenman:
I couldn't find any code that depended on the character branch (nor can I imagine a good use case), but I figured that since the old code handled it, it would be better if your changes also do. This is probably something that should be checked for and removed at some point, but better not to do that so short before a release.

R/common.R
Kucharssim: good to know!

vandenman (approved):

PR #80, opened 2022-09-20
fixing computed column by passing pointers, and more

JorisGoosen:
Make sure pointers are passed to jaspBase with relevant functions for computed columns etc

Part of the fix for https://github.com/jasp-stats/jasp-test-release/issues/1967

Also couldnt resist removing all the now no longer necessary jaspNativeUtf8??? functions etc

### DIFF
--- a/configure
+++ b/configure
@@ -110,6 +110,5 @@ if [ "${JASP_R_INTERFACE_LIBRARY}" != "" ]; then
 fi
 
 sed -e "s|@cppflags@|${PKG_CXXFLAGS}|" -e "s|@src_sources@|${SRC_SOURCES}|" -e "s|@jaspColumnEncoder_sources@|${JASPCOLUMNENCODER_SOURCES}|" src/Makevars.in > src/Makevars
-#sed -e "s|@cppflags@|${PKG_CXXFLAGS_EXPANDED}|" -e "s|@src_sources@|${SRC_SOURCES}|" -e "s|@jaspColumnEncoder_sources@|${JASPCOLUMNENCODER_SOURCES}|" -e "s|@jasp_r_interface@|${JASP_R_INTERFACE_LIBRARY_VALUE}|" src/Makevars.in > src/Makevars
 
 exit 0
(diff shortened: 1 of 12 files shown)


src/jaspObject.h
```diff
@@ -1,29 +1,21 @@
-#pragma once
+#ifndef JASPOBJECT_MANUAL_INCLUDE_GUARD
+#define JASPOBJECT_MANUAL_INCLUDE_GUARD
```
vandenman, line 2: Just for my own understanding, is there anything special about this include guard? Just wondering because its name would suggest that and it's not `JASPOBJECT_H` like QT would generate.

configure
```diff
@@ -104,6 +104,12 @@ fi
 SRC_SOURCES="$(cd src/ && ls *.cpp | tr '\n' ' ')"
 JASPCOLUMNENCODER_SOURCES="$(cd src/ && ls ${JASPCOLUMNENCODER_DIR}/*.cpp | tr '\n' ' ')"
 
+#add the define if JASP_R_INTERFACE_LIBRARY env var says anything at all
+if [ "${JASP_R_INTERFACE_LIBRARY}" != "" ]; then
+	PKG_CXXFLAGS=-DJASP_R_INTERFACE_LIBRARY\ ${PKG_CXXFLAGS}
+fi
+
 sed -e "s|@cppflags@|${PKG_CXXFLAGS}|" -e "s|@src_sources@|${SRC_SOURCES}|" -e "s|@jaspColumnEncoder_sources@|${JASPCOLUMNENCODER_SOURCES}|" src/Makevars.in > src/Makevars
+#sed -e "s|@cppflags@|${PKG_CXXFLAGS_EXPANDED}|" -e "s|@src_sources@|${SRC_SOURCES}|" -e "s|@jaspColumnEncoder_sources@|${JASPCOLUMNENCODER_SOURCES}|" -e "s|@jasp_r_interface@|${JASP_R_INTERFACE_LIBRARY_VALUE}|" src/Makevars.in > src/Makevars
```
vandenman, line 113:
```suggestion
```
I know I added this, but it's probably better to delete it :grin:

vandenman (approved):
looks good to me, but let's run the integration tests to make sure

boutinb (approved):

src/jaspObject.h
JorisGoosen: No nothing special. It is the same as the `#pragma once` or whatever qt generates. I was running into some duplicates during linking. 

But that was solved by removing `#include "jaspResults.h"` and this is functionally the same anyway

PR #81, opened 2022-09-22
contr functions from kknn

Kucharssim:
fixes https://github.com/jasp-stats/jasp-test-release/issues/2000

for some reason simply adding the function names to the `doNotRemoveAnyOfThese` here:
https://github.com/jasp-stats/jaspBase/blob/5e839aa73965233894953b16d509cc778c435f88/R/memoryMaintenance.R#L24

did not work, but importing these functions into `jaspBase` did... this means this PR is adding `kknn` as a dependency to `jaspBase`, which less than great. 

But maybe @JorisGoosen had a different idea for the fix and I misunderstood...


JorisGoosen:
Ok my idea of a fix doesnt seem to work but depending on `kknn` is also bad....

@vandenman ?

vandenman:
I haven't looked into this but since we're already using a fork of kknn (vandenman/kknn), can't we fix this there?

Kucharssim:
We could in theory just copy paste the functions from `kknn`, so that `jaspBase` does not have to depend on it. The downside is that it's just stupid ugly solution (and we need check how to include it such that we don't break `kknn`'s license)

JorisGoosen:
> I haven't looked into this but since we're already using a fork of kknn (vandenman/kknn), can't we fix this there?

This sounds like my favourite solution

Also, could we not make dummy functions in `jaspBase` that simply call `kknn` if available and not otherwise.
Then we could put `kknn` in suggests?

Kucharssim:
> I haven't looked into this but since we're already using a fork of kknn (vandenman/kknn), can't we fix this there?

Well it's possible, but since `kknn` uses these functions by setting them as values of `contrasts` in `options()`, it's kind of pain to make sure that you don't miss anything as the functions can be in theory called implicitly from anywhere.

Kucharssim:
lol, well the easiest option is to fix it directly in `jaspMachineLearning`, no need for fiddling with `jaspBase` or `kknn` apparently...

JorisGoosen (approved):

Kucharssim:
Yeah the other fixes did not really work

PR #84, opened 2022-10-04
Add position in placeholders inside gettextf

boutinb:
See issue https://github.com/jasp-stats/INTERNAL-jasp/issues/2125


po/R-zh_Hant.po
```diff
@@ -97,20 +97,20 @@ msgstr ""
 msgid "Infinity found in %s"
 msgstr "%s 出現無限值"
 
-msgid "Number of factor levels is %s in %s"
-msgstr "在 %s 的因素個數為 %s"
+msgid "Number of factor levels is %1$s in %2$s"
+msgstr "在 %1$s 的因素個數為 %2$s"
```
shun2wang, line 101: for example, the words order is ignored by the translator , and we seem to need an additional message to them by email after all the work is completed,Or on the eve of our next version release？

shun2wang (approved):

po/R-zh_Hant.po
boutinb: If the translator does not use the right placeholder, WebLate should already emit a warning. Also the GitHub translation action generates an error. I check these errors from time to time, and I either solve the problem directly in WebLate, or I send an email to the translator if I have some doubts.

shun2wang: sorry for @boutinb 😭，I forgot a little just now but we missing some strings here:
`
line 73: stop(gettextf("Could not find message for class %s and type %s", as.character(class), as.character(type)))`

we need anther PR...

PR #88, opened 2022-10-07
New method for module installation

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1887
part of https://github.com/jasp-stats/INTERNAL-jasp/issues/1782

See https://github.com/jasp-stats/jasp-desktop/pull/4884 for an explanation.

This needs to be tested extensively.


R/checkSums.R
```diff
@@ -1,19 +1,30 @@
...
+    list.files(file.path(modulePkg, "R"),       recursive = FALSE, full.names = TRUE, pattern = "\\.R$"),
+    list.files(modulePkg,                       recursive = FALSE, full.names = TRUE, pattern = "renv\\.lock")
   )
-  newMd5Sums <- tools::md5sum(srcFiles)
-  newMd5Sums
+
+  if (includeQML)
+    srcFiles <- c(
+      srcFiles,
+      list.files(file.path(modulePkg, "inst"),  recursive = TRUE, full.names = TRUE, pattern = "\\.(qml|po|svg|png|jpg|md)$"),
```
JorisGoosen, line 13: this covers Upgrades and Description.qml as well right?

R/moduleInstall.R
```diff
@@ -361,15 +364,648 @@ setupRenv <- function(moduleLibrary, modulePkg) {
 
   Sys.setenv("RENV_PATHS_LIBRARY" = moduleLibrary)
 
-  print("Using the following paths:")
+  cat("Using the following paths:\n")
```
JorisGoosen, line 367: Now we can't see this in the log anymore when running a dynamic module install right?

Why would we want that?

R/moduleInstall.R
```diff
@@ -361,15 +364,648 @@ setupRenv <- function(moduleLibrary, modulePkg) {
...
+
+    if (any(df$identical)) # ensure the folders also exist
+      df$identical[df$identical] <- df$identical[df$identical] & dir.exists(file.path(moduleLibrary, jaspPkgs[df$identical]))
+
+    # if the module didn't change then we can reuse the lockfile, otherwise we reinstall everything from scratch
+    reusingLockfile <- df[[moduleName, "identical"]]
+
+    if (reusingLockfile) {
+
+      if (verbose >= 1) cat("Package hash in lockfile identical to local folder, reusing already existing lockfile\n")
```
JorisGoosen, line 523: I would put the `if (verbose >= 1)` before `if(reusingLockfile)`

R/moduleInstall.R
```diff
@@ -361,15 +364,648 @@ setupRenv <- function(moduleLibrary, modulePkg) {
...
+#' Possible values for \code{updatePackages} are "true", "false", "daily", "triweekly", "biweekly", "weekly", "fortnightly", "monthly",
+#' "monday", "tuesday", ... "sunday", or a combination of weekdays, "monday;wednesday;friday".
+#' Anything value that is not understood is interpreted as FALSE (with a warning).
+#' The date of the previous update is determined by looking at the last date where \code{lockfilePath} was modified.
+#'
+#' @rdname installModuleNew
+#'
+#' @return returns \code{NULL}.
+#'
+installModuleNew <- function(
```
JorisGoosen, line 443: This function is so large I think it would be good to have some comments in between breaking up the flow describing what each section does. Like "Here we are going to scan the lockfile" or something.
It would help me, and probably future you, out a lot.

Splitting it up in some descriptively named functions would also be a good way to solve this but I can imagine that would be a bit painful so I wont ask for that.

JorisGoosen (changes requested):

JorisGoosen:
well this is one massive set of changes :o

Ive skimmed through it now only. Ill also try to build this together with jasp-desktop and see whether it works.
Im also going to look into installing dyanmic modules afterwards because that also has to keep working (insofar as it does)

R/checkSums.R
vandenman: yes, those should be included. Even helpfiles with extension .md are included

R/moduleInstall.R
vandenman: This should be visible in both? This code is, for now, intentionally not used when installing a dynamic module, so I'm not 100% sure though.

R/moduleInstall.R
vandenman: Will do!

vandenman:
>  Im also going to look into installing dyanmic modules afterwards because that also has to keep working (insofar as it does)

This is why I didn't delete any of the old code (yet). Dynamic modules still uses that. This only concerns the build process. I can also test this with dynamic modules immediately if you prefer, but I figured the current changes have gotten large enough :sweat_smile:

JorisGoosen:
> > Im also going to look into installing dyanmic modules afterwards because that also has to keep working (insofar as it does)
> 
> This is why I didn't delete any of the old code (yet). Dynamic modules still uses that. This only concerns the build process. I can also test this with dynamic modules immediately if you prefer, but I figured the current changes have gotten large enough 😅

Yes which is good, the keeping of the old code that is, but there could still be problems with the links in the libraries in the build/install-folder. What could happen is that jaspBase or jaspModules library links to `renv-cache` get changed by `renv` to point to stuff in development-module-libraries or module installs... So we also need to test with that because I really dont want that stuff to break again. Maybe we could even make a unittest for that?

R/moduleInstall.R
JorisGoosen: But I assume that later we will reuse this code also for dynamic modules right?

Those are going to be exactly the same after all, as the ones in jasp-desktop.

But im not sure what the gain is from changing them from `print` anyway?

vandenman:
superseded

PR #91, opened 2022-10-18
add `toRObject` from previous branch

vandenman:
This was still lying around on my fork of jaspResults, so I figured I could make it into a PR.

This adds the method `toRObject` to all jasp objects that are subclasses of `jaspOutputObjR`. This means that subclasses of `jaspObjR` (e.g., `jaspState`) do not have this method, but I'm also not sure if they should.

Here an example:

``` r
rm(list = ls())
library(jaspBase)

# otherwise jaspPlot crashes badly (should be fixed another time)
jaspResultsCPP        <- jaspBase:::loadJaspResults("a name")
jaspResultsCPP$title  <- "a title"
jaspResults           <- jaspBase:::jaspResultsR$new(jaspResultsCPP)

dummyTable <- function(title = "hola") {
  tb <- createJaspTable(title = title, rowNames = letters[1:4]) # intentionally one too much
  tb$addColumnInfo("model", title =  "Model",   type = "string")
  tb$addColumnInfo("lower", title =  "Lower",   type = "number", overtitle = "95% Confidence interval")
  tb$addColumnInfo("upper", title =  "Upper",   type = "number", overtitle = "95% Confidence interval")
  tb$addColumnInfo("pval",  title =  "p-value", type = "pvalue")

  set.seed(123)
  tb$setData(data.frame(
    model = paste0("H", 0:2),
    lower = rnorm(3),
    upper = rnorm(3),
    pval  = runif(3)
  ))

  # tb$setRowNames(letters[1:4])

  tb$addFootnote("first footnote", symbol = "symbol a")
  tb$addFootnote("2nd   footnote", symbol = "symbol b", colNames = "lower")
  tb$addFootnote("derde footnote", symbol = "symbol c", colNames = c("lower", "upper"))
  tb$addFootnote("||||  footnote", symbol = "symbol d", colNames = c("lower", "upper"), rowNames = c("a", "c"))
  tb$addFootnote("funf  footnote", symbol = "symbol e", rowNames = "a")

  tb
}

dummyPlot <- function() {
  plt <- createJaspPlot(title = "I'm a plot!")
  set.seed(42)
  plt$plotObject <- ggplot2::qplot(rnorm(10), rnorm(10))
  plt
}

dummyContainer <- function() {
  cont <- createJaspContainer(title = "containerino")
  cont[["tablio"]] <- dummyTable()
  cont[["plotio"]] <- dummyPlot()
  cont[["table2"]] <- dummyTable("Also a table")
  cont[["html"]]   <- createJaspHtml(text = "some html", title = "that's not hidden")
  cont[["state"]]  <- createJaspState(object = list(1))
  cont
}

ct <- dummyContainer()
oo <- ct$toRObject()
oo
#> $hola
#>   model      lower      upper      pval
#> a    H0 -0.5604756 0.07050839 0.6775706
#> b    H1 -0.2301775 0.12928774 0.5726334
#> c    H2  1.5587083 1.71506499 0.1029247
#> 
#> $`I'm a plot!`
#> $`I'm a plot!`
#> 
#> attr(,"title")
#> [1] "I'm a plot!"
#> attr(,"class")
#> [1] "jaspPlotWrapper" "jaspWrapper"    
#> attr(,"jaspObjectEnvironment")
#> <environment: 0x55b37223e640>
#> 
#> $`Also a table`
#>   model      lower      upper      pval
#> a    H0 -0.5604756 0.07050839 0.6775706
#> b    H1 -0.2301775 0.12928774 0.5726334
#> c    H2  1.5587083 1.71506499 0.1029247
#> 
#> $`that's not hidden`
#> $rawtext
#> [1] "some html"
#> 
#> $text
#> [1] "<p>some html </p>"
#> 
#> $class
#> [1] ""
#> 
#> $maxWidth
#> [1] "15cm"
#> 
#> $elementType
#> [1] "p"
#> 
#> attr(,"title")
#> [1] "that's not hidden"
#> attr(,"class")
#> [1] "jaspHtmlWrapper" "jaspWrapper"    
#> attr(,"jaspObjectEnvironment")
#> <environment: 0x55b372475170>
#> 
#> attr(,"class")
#> [1] "jaspContainerWrapper" "jaspWrapper"         
#> attr(,"title")
#> [1] "containerino"
#> attr(,"jaspObjectEnvironment")
#> <environment: 0x55b3724c8bb0>
class(oo)
#> [1] "jaspContainerWrapper" "jaspWrapper"
names(oo)
#> [1] "hola"              "I'm a plot!"       "Also a table"     
#> [4] "that's not hidden"
str(oo, max.level = 1)
#> List of 4
#>  $ hola             :Classes 'jaspTableWrapper', 'jaspWrapper' and 'data.frame': 3 obs. of  4 variables:
#>   ..- attr(*, "footnotes")=List of 5
#>   ..- attr(*, "title")= chr "hola"
#>   ..- attr(*, "jaspObjectEnvironment")=<environment: 0x55b3721b7f68> 
#>  $ I'm a plot!      :List of 1
#>   ..- attr(*, "title")= chr "I'm a plot!"
#>   ..- attr(*, "class")= chr [1:2] "jaspPlotWrapper" "jaspWrapper"
#>   ..- attr(*, "jaspObjectEnvironment")=<environment: 0x55b37223e640> 
#>  $ Also a table     :Classes 'jaspTableWrapper', 'jaspWrapper' and 'data.frame': 3 obs. of  4 variables:
#>   ..- attr(*, "footnotes")=List of 5
#>   ..- attr(*, "title")= chr "Also a table"
#>   ..- attr(*, "jaspObjectEnvironment")=<environment: 0x55b372315818> 
#>  $ that's not hidden:List of 5
#>   ..- attr(*, "title")= chr "that's not hidden"
#>   ..- attr(*, "class")= chr [1:2] "jaspHtmlWrapper" "jaspWrapper"
#>   ..- attr(*, "jaspObjectEnvironment")=<environment: 0x55b372475170> 
#>  - attr(*, "class")= chr [1:2] "jaspContainerWrapper" "jaspWrapper"
#>  - attr(*, "title")= chr "containerino"
#>  - attr(*, "jaspObjectEnvironment")=<environment: 0x55b3724c8bb0>
```

One perhaps weird element is that each jaspObject contains the attribute "jaspObjectEnvironment" with an environment containing the original S4 jaspObject. By keeping a reference to this object, we can then later forward the print function to the print function of a fancy jasp object, so that we do not need to write two fancy print functions. We may want to try out different ways to organize this reference at some point.

@Kucharssim any feedback on the returned R structure?

@JorisGoosen any feedback on specifically
  - how can we make `toRObject()` const? If that's even possible? I guess there is a reason that `jaspContainer::at` isn't const either...
  - how to complete `Rcpp::List footnotes::toRObject()`? I don't really understand what can be in `tableFields`. The data are `std::set<Json::Value>`, but as far as I know this can only be a vector of strings? Either way, it would be good to know all possible options so that we can safely convert it to an R object.


Kucharssim:
>  This means that subclasses of jaspObjR (e.g., jaspState) do not have this method, but I'm also not sure if they should.

I think of this functionality as exposing to R the stuff that users see in the JASP output, so doing this only for `jaspOutputObjR` sounds very reasonable to me.

Looks and works nice in R, couple of comments/questions:

- Should the tables respect the column type info definition in `$addColumnInfo()`? For example,

```r
table <- createJaspTable("A table")
table$addColumnInfo("x", type = "number")
table$setData(data.frame(x=1:5))
table$toRObject()$x |> class()
#> [1] "integer" # I would expect "numeric"
```
In 99% of cases this probably does not matter anyway, so it's a minor issue, but just wondering.

- Would it make sense to also export the column `title`s? It makes sense to use the `name` for the column names (even though that will cause us some pain because those names are not used consistently across JASP), but I imagine the `titles` could be helpful as well - perhaps added to the object as one of the attributes just in case the end user wants to do something with it?

- Should there be a way for the users to access the `ggplot` objects? Ultimately I would imagine users would like to do something like

```r
results <- anova(...)
myPlot <- results[["Some JASP Plot"]]
myPlot <- myPlot + ggplot2::xlab("I want to change something about the plot returned by JASP!!!")
```

but that would not work with the current setup obviously.

- What's your plan for the new classes like `jaspPlotWrapper`, etc? Do we plan to implement some methods for these things? You mention pretty printing, so that could be one of the use cases (even though the tables print pretty already because they are `data.frame`s after all).

- My R session crashes if I try to convert an empty table:

```r
table <- createJaspTable()
table$toRObject()
```

vandenman:
> Should the tables respect the column type info definition in $addColumnInfo()? For example,

I think they should. This is just how jaspResults currently interprets that data, so perhaps we need to change something there.

> Would it make sense to also export the column titles? It makes sense to use the name for the column names (even though that will cause us some pain because those names are not used consistently across JASP), but I imagine the titles could be helpful as well - perhaps added to the object as one of the attributes just in case the end user wants to do something with it?

I can put those in the table as another attribute. I decided not to use them as column names, since the column titles often contains spaces or other characters that R normally disallows in dataframes. 

> Should there be a way for the users to access the ggplot objects? Ultimately I would imagine users would like to do something like

I'll check again, but I thought you could? In the example, `oo[["I'm a plot!"]]` is just a classed list of a ggplot. You can just add a new theme iirc.

> What's your plan for the new classes like jaspPlotWrapper, etc? Do we plan to implement some methods for these things? You mention pretty printing, so that could be one of the use cases (even though the tables print pretty already because they are data.frames after all).

Exactly. In the future there could be print and summary methods. For example, we could do

```r
print.jaspWrapper <- function(x, ...) {
  attr(obj, "jaspObjectEnvironment")$jaspObject$print()
}
```

```r
summary.jaspTableWrapper <- function(x, ...) {
  # specific stuff for tables
}
```

> My R session crashes if I try to convert an empty table:

uh oh :sweat_smile: I'll take a look.

vandenman:
> But I don't think it warrants another mammoth project similar to options renaming, or does it?

No, R handles the weird names fine and from now on we should just make sure that the tables do not get too funky internal names.

> For the pretty printing of tables I would imagine we could show both the title and name of each column, 

There are multiple ways to do that. My current plan was to only print the internal names if some option is set, see for example the comment [here](https://github.com/vandenman/jaspResults/blob/db8753bfef7ffd9c7ea045e9c45c15381e4c7183/src/jaspTable.cpp#L771-L788). By default, the printed table is an ascii representation of what is shown inside Jasp.

> Could we do it without wrapping the ggplot in a list though?

We can, but then it might become weird if we add custom classes and attributes to a ggplot and overwrite the default print method. Perhaps it makes more sense if we name the element? So it would be ``oo$`I'm a plot!`$plotObject`` or so? I can also imagine that we want to add other information (e.g., width & height) in there, although those could also be added as attributes.

Kucharssim:
> We can, but then it might become weird if we add custom classes and attributes to a ggplot and overwrite the default print method. Perhaps it makes more sense if we name the element? So it would be oo$`I'm a plot!`$plotObject or so? I can also imagine that we want to add other information (e.g., width & height) in there, although those could also be added as attributes.

True, good point. One alternative I can think of is to implement a `+` method for `jaspPlotWrapper` objects, which seems to work on this example

```r
dummyPlot <- function() {
  out <- list(
    plotObject = ggplot2::ggplot(data.frame(x = rnorm(100)), ggplot2::aes(x = x)) + ggplot2::geom_histogram()
  )
  
  class(out) <- c("jaspPlotWrapper")
  return(out)
}

`+.jaspPlotWrapper` <- function(e1, e2) {
  e1$plotObject <- e1$plotObject + e2
  return(e1)
}

dummyPlot() + ggplot2::xlab("something")
```

Assuming we won't need the binary operator for other manipulations of the `jaspPlotWrapper` objects, this could be a way how to ensure that we don't need to modify the `ggplot` objects but people can add `ggplot` statements to it *as it it was one*.

R/jaspObjectWrapperMethods.R
```diff
@@ -0,0 +1,5 @@
+#' @export
```
JorisGoosen, line 1: Comment please?

src/jaspTable.cpp
```diff
@@ -1072,6 +1154,28 @@ void footnotes::convertToJSONOrdered(std::map<std::string, size_t> rowNames, std
...
+
+	for (const auto & textRest : _data)
+		for(const auto & symbolRest : textRest.second)
+			for(const tableFields & fields : symbolRest.second)
+			{
+				Rcpp::List note = Rcpp::List::create(
+					Rcpp::Named("text")		= textRest.first,
+					Rcpp::Named("symbol")	= symbolRest.first
+//					TODO: I do not understand the data in here, or how to convert it to R...
+//					Rcpp::Named("rows")		= fields.rowsToJSON(),
```
JorisGoosen, line 1171: We can just convert to a Rcpp::List and be done with it no?
The information is available then and that should be enough for now.

JorisGoosen (approved):

PR #92, opened 2022-10-26
Add dependencies on nested options

vandenman:
Required for some changes I'm making to Jags and Quality Control.

@Kucharssim I'd appreciate your feedback on the R-interface!

For any jaspObject, you can do
```r
obj$dependOn(
  # previously available
  options=...
  optionsFromObject=...
  optionContainsValue...
  # new in this PR
  nestedOptions=...
  nestedOptionsContainsValue=...
)
```
For `nestedOptions`, either pass a single character vector, for example `nestedOptions = c("model", "3", "checkbox")` to add a dependency on `options[["model"]][[3]][["checkbox"]]`, or use a list of character vectors to specify multiple nested dependencies:
```r
obj$dependOn(nestedOptions = list(
  c("model", "3", "checkboxA"),
  c("model", "3", "checkboxB"),
))
```

For `nestedOptionsContainsValue` the idea is to pass a list of key-value pairs where odd indices are interpreted as keys and even indices as values. For example:
```r
obj$dependOn(nestedOptionsContainsValue = list(
  c("model", "3", "variableField"), # key
  "contNormal",                     # value
  c("model", "3", "variableField"), # key
  "contGamma"                       # value
)
```
For readability, we should recommend that people write this in two aligned columns:
```r
obj$dependOn(nestedOptionsContainsValue = list(
  # keys                            values
  c("model", "3", "variableField"), "contNormal",
  c("model", "3", "variableField"), "contGamma"
))
```
this interface is documented as comments in zzzWrappers.R

### DIFF
--- a/src/jaspObject.cpp
+++ b/src/jaspObject.cpp
@@ -126,6 +126,29 @@ void jaspObject::removeChild(jaspObject * child)
 	child->parent = NULL;
 }
 
+Json::Value jaspObject::getObjectFromNestedOption(std::vector<std::string> nestedKey, Json::Value ifNotFound) const
+{
+	Json::Value obj = currentOptions;
+	for (const auto& key: nestedKey)
+	{
+		// NOTE: this fails if we have options which are an array where some elements are named, but not all.
+		// I think that would violate the Json spec for arrays, but I'm not 100% sure.
+		if (obj.isArray() && !key.empty() && std::all_of(key.begin(), key.end(), ::isdigit))
+		{
+			int index = stoi(key);
+			obj = obj.get(index, Json::nullValue);
+		}
+		else
+		{
+			obj = obj.get(key, Json::nullValue);
+		}
+
+		if (obj.isNull())
+			return ifNotFound;
+	}
+	return obj;
+}
+
 void jaspObject::finalized()
 {
 	//std::cout << "jaspObject::finalized() called on "<<objectTitleString()<<" " << (_finalizedAlready ? "again!" :"") << "\n" << std::flush;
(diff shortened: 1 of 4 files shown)


Kucharssim:
> For nestedOptions, either pass a single character vector, for example nestedOptions = c("model", "3", "checkbox") to add a dependency on options[["model"]][[3]][["checkbox"]]

How does it know to index by `3` and not `"3"`? Can I break it if I use a named list with names c("1", "3", "2")?

> For readability, we should recommend that people write this in two aligned columns:

Could we enforce this by requiring a list of lists instead?

```r
obj$dependOn(nestedOptionsContainsValue = list(
  list(key = c("model", "3", "variableField"), value = "contNormal"),
  list(key = c("model", "3", "variableField"), value = "contGamma")
))
```

Yeah it's more verbose, but more readable, I think.

Further, I am not sure whether it's not a typo in your examples, but would it make sense to collapse specification in case the key stays the same? Instead of 

```r
  list(key = c("model", "3", "variableField"), value = "contNormal"),
  list(key = c("model", "3", "variableField"), value = "contGamma")
```

have 

```r 
list(key = c("model", "3", "variableField"), value = c("contNormal", "contGamma")
```

vandenman:
> How does it know to index by 3 and not "3"? 

Because I think/ assume that JSON arrays cannot have names. So if some nested option is an array and the key is convertible to a number, the code tries to use it as an index.

> Could we enforce this by requiring a list of lists instead?

Sure. I'll probably only change this on the R side, internally we loop over the list anyway and do `private$jaspObject$setNestedOptionMustContainDependency(key, value)`.

> would it make sense to collapse specification in case the key stays the same?

Yes, that would make sense, I'll see if I can make that work.

src/jaspObject.cpp
```diff
@@ -490,32 +517,17 @@ bool jaspObject::checkDependencies(Json::Value currentOptions)
...
 		for(auto & keyval : _optionMustContain)
-		{
-			bool foundIt = false;
-
-			for(auto & contains : currentOptions.get(keyval.first, Json::arrayValue))
-				if(contains == keyval.second)
-					foundIt = true;
-
-			if(!foundIt)
+			if (!isJsonSubArray(keyval.second, currentOptions.get(keyval.first, Json::arrayValue)))
```
vandenman, line 520: @JorisGoosen I don't think the old code ever worked properly here, so I took the liberty of fixing it. Also, see https://github.com/jasp-stats/jaspTestModule/pull/41 for testing. When I tried the old comparison it didn't work unless I ordered things such that `contains == keyval.second` (but it should also work when `keyval.second ∈ contains`).

JorisGoosen: Aah the `keyval.second ∈ contains` was something I never intended so that didnt work. But I can see how it would be useful yeah.

src/jaspObject.cpp
```diff
@@ -126,6 +126,96 @@ void jaspObject::removeChild(jaspObject * child)
 	child->parent = NULL;
 }
 
+Json::Value jaspObject::getObjectFromNestedOption(std::vector<std::string> nestedKey, Json::Value ifNotFound) const
+{
+	Json::Value obj = currentOptions;
+	for (const auto& key: nestedKey)
+	{
+		// NOTE: this fails if we have options which are an array where some elements are named, but not all.
+		// I think that would violate the Json spec for arrays, but I'm not 100% sure.
```
JorisGoosen, line 135: I dont think we can have an array with names in C++, `Json::array != Json::object` luckily

vandenman: good to know! I wasn't sure because this is possible in R.

JorisGoosen (approved):
Nice work!

PR #95, opened 2022-11-07
add a check for output of system2(c(renv:::R(),"CMD","config","CC"),stdout=TRUE,stderr=TRUE)

JorisGoosen:
This helps avoid some annoying errors on macOS where people updated the OS and have to reinstall the xcode commandline tools but are unaware. And the error message is something like:
`Error in if (eval(cond, envir = environment(dot))) return(eval(expr, envir = environment(dot))): the condition has length > 1`

### DIFF
--- a/R/moduleInstall.R
+++ b/R/moduleInstall.R
@@ -39,14 +39,12 @@ installJaspModule <- function(modulePkg, libPathsToUse, moduleLibrary, repos, on
   cmdConfigCC <- system2(c(renv:::R(),"CMD","config","CC"),stdout=TRUE,stderr=TRUE)
   if(length(cmdConfigCC) > 1)
     stop(
-      paste0(
 "R CMD config CC returns more than 1 line, this will break renv and thus your install. 
 Most likely you are on mac and you should run `xcode-select --install` in a terminal. 
 If that doesn't help or you aren't on a mac: feel free to open an issue at https://github.com/jasp-stats/jasp-issues/issues/new/choose
   
 The output was: 
-"     ), 
-      paste(cmdConfigCC, sep="", collapse="\n"))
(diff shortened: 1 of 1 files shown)


JorisGoosen:
Ooh yeah so I cant really test this until I update xcode/macOS again apparently. But I think the code is sound.

R/moduleInstall.R
```diff
@@ -34,6 +34,20 @@ installJaspModule <- function(modulePkg, libPathsToUse, moduleLibrary, repos, on
...
+  if(length(cmdConfigCC) > 1)
+    stop(
+      paste0(
+"R CMD config CC returns more than 1 line, this will break renv and thus your install. 
+Most likely you are on mac and you should run `xcode-select --install` in a terminal. 
+If that doesn't help or you aren't on a mac: feel free to open an issue at https://github.com/jasp-stats/jasp-issues/issues/new/choose
+  
+The output was: 
+"     ), 
+      paste(cmdConfigCC, sep="", collapse="\n"))
```
vandenman, line 49:
```suggestion
    stop(
"R CMD config CC returns more than 1 line, this will break renv and thus your install. 
Most likely you are on mac and you should run `xcode-select --install` in a terminal. 
If that doesn't help or you aren't on a mac: feel free to open an issue at https://github.com/jasp-stats/jasp-issues/issues/new/choose
  
The output was: 
", paste0(cmdConfigCC, collapse="\n"), domain = NA)
```
- `domain = NA` avoid this being translated. 
- `stop` pastes all unnamed arguments together anyway so you don't need the first `paste`.
- `paste0(..., sep = "") === paste0(...)`.

vandenman:
one minor comment, other than that everything looks fine

PR #96, opened 2022-11-08
implement setOrRetrieve, jaspDeps, and isRecomputed

vandenman:
as discussed in Slack.


R/setOrRetrieve.R
```diff
@@ -0,0 +1,108 @@
...
+
+saveHashOfJaspObject <- function(x) {
+  .internal[["recomputedHashtab"]][[x]] <- 0L
+}
+
+setRecomputed <- function(x) {
+  .internal[["lastRecomputed"]] <- x
+}
+
+#' Test if a jaspObject was recomputed of not.
```
Kucharssim, line 14:
```suggestion
#' Test if a jaspObject was recomputed or not.
```

R/setOrRetrieve.R
```diff
@@ -0,0 +1,108 @@
...
+#' @param x a jaspObject, or missing in which case the last created or retrieved jaspObject is considered.
+#'
+#' @details Note that this function only works when \code{\link{\%setOrRetrieve\%}} is used to set or retrieve a jaspObject.
+#'
+#' @export
```
Kucharssim, line 19: I wonder if `isRecomputed` and `%setOrRetrieve%` could be put under one help file. This makes it clearer that they are to be used in tandem, which is also shown in the example file.

R/setOrRetrieve.R
```diff
@@ -0,0 +1,108 @@
...
+}
+
+setRecomputed <- function(x) {
+  .internal[["lastRecomputed"]] <- x
+}
+
+#' Test if a jaspObject was recomputed of not.
+#' @param x a jaspObject, or missing in which case the last created or retrieved jaspObject is considered.
+#'
+#' @details Note that this function only works when \code{\link{\%setOrRetrieve\%}} is used to set or retrieve a jaspObject.
```
Kucharssim, line 17: any reason to use `\code{\link{\%setOrRetrieve\%}}` instead of `[%setOrRetrieve]`? The latter is more readable to me.

R/setOrRetrieve.R
```diff
@@ -0,0 +1,108 @@
+.internal <- list2env(list(
+  recomputedHashtab = hashtab(type = "address", NULL), # unclear if "address" is the correct choice here
```
Kucharssim, line 2: This is indeed unclear to me too: What's your reason to choose `address` here? The docs say that `identical` is almost always the better choice, but here we are in a highly unusual setting, so I am willing to bet on your reasoning...

R/setOrRetrieve.R
```diff
@@ -0,0 +1,108 @@
...
+#' If the `rhs` does not return a jaspObject, an error is thrown.
+#'
+#' @rdname setOrRetrieve
+#' @example inst/examples/ex-setOrRetrieve.R
+#' @export
+`%setOrRetrieve%` <- function(lhs, rhs) {
+
+  exprLhs <- substitute(lhs) # need to do this before evaluating lhs
+  if (!is.null(lhs)) {
+
```
Kucharssim, line 81: hmm, could we check here if `lhs` (or the *parent* of it) is a jaspObject? Because currently I can do

```r
a <- 1
jaspBase:::.internal$recomputedHashtab[[a]]
# [1] NULL
a %setOrRetrieve% 2
# [1] 1
jaspBase:::.internal$recomputedHashtab[[a]]
# [1] 0
```
which does not seem ideal...

vandenman: yes, I'll try to do that.

R/setOrRetrieve.R
vandenman: my reasoning is that we're using R6 objects as hash, which are environments and are always unique (copies need to be made explicitly, but we basically don't do that). Perhaps this is a premature optimization though.

R/setOrRetrieve.R
vandenman: `[%setOrRetrieve]` does not create a link to the helpfile or does it? Either way, I merged the helpfiles so I replaced everything by \`%setOrRetrieve%\`.

R/setOrRetrieve.R
vandenman: done

R/setOrRetrieve.R
Kucharssim: > [%setOrRetrieve] does not create a link to the helpfile or does it? 

it does, if it's referencing a function in the same package (otherwise you need to prefix with `package::`). The link won't show up properly if you render the help files using `devtools::document()`, but if you install normally it should work

Kucharssim (approved):

PR #98, opened 2022-11-10
Only columnencoder.cpp and columntype must be compiled

boutinb:
As jaspColumnEncoder submodule will soon disappear, and be added directly in Common (which will be also a submodule), jaspBase should not compile all classes in Common but only what is needed for columnencoder.

This fix is needed for https://github.com/jasp-stats/jasp-desktop/pull/4914

This fix is needed because jaspBase generates error when compiling all files from Common: apparently R uses gnu++14 (JASP uses c++17), and some classes from Common includes <filesystem> which generates error in c++14.

### DIFF
--- a/configure
+++ b/configure
@@ -102,7 +102,7 @@ location of the parent directory (e.g., the default is '-I\"../inst/include/jasp
 fi
 
 SRC_SOURCES="$(cd src/ && ls *.cpp | tr '\n' ' ')"
-JASPCOLUMNENCODER_SOURCES="$(cd src/ && ls ${JASPCOLUMNENCODER_DIR}/*.cpp | tr '\n' ' ')"
+JASPCOLUMNENCODER_SOURCES="${JASPCOLUMNENCODER_DIR}/columnencoder.cpp ${JASPCOLUMNENCODER_DIR}/columntype.cpp"
 
 #add the define if JASP_R_INTERFACE_LIBRARY env var says anything at all
 if [ "${JASP_R_INTERFACE_LIBRARY}" != "" ]; then
--- a/configure.win
+++ b/configure.win
@@ -87,7 +87,7 @@ location of the parent directory (e.g., the default is '-I\"../inst/include/jasp
 fi
 
 SRC_SOURCES="$(cd src/ && ls *.cpp | tr '\n' ' ')"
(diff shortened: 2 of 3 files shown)


vandenman:
perhaps we should also adjust https://github.com/jasp-stats/jaspBase/blob/e1509da5bad7c7acb3b43b3dd9cc1dbe462edb31/src/Makevars.in#L1-L3

to specify c++17?

boutinb:
> to specify c++17?

I can do that indeed. But this fix is anyway a good thing, since jaspBase does not need to compile all the other classes.

boutinb:
I have changed to c++ 17, but if I compile it, I still get:
`clang++ -arch arm64 -std=gnu++14`
And nothing about c++ 11 or C++ 17....
But yeah, it's still works.

vandenman:
With these changes:
```diff
-CXX_STD				= CXX11 
+CXX_STD				= CXX20
PKG_CXXFLAGS		= @cppflags@
-CXX11STD			= -std=c++11
+CXX20STD			= -std=c++2a
```
and then run `R CMD INSTALL --preclean --no-multiarch --with-keep.source jaspBase` I get:
image

If it's okay with you, I'd prefer to go to C++20. In particular, the [format library](https://www.cppstories.com/2022/custom-stdformat-cpp20/) will make my life much easier when creating pretty print functions for e.g., jaspTables and other objects.

vandenman (approved):
compiles fine! Didn't check on windows though

vandenman:
So that doesn't work for me :sweat: the only change I need is this:

```diff
-CXX_STD				= CXX11 
+CXX_STD				= CXX20
```
and perhaps we should also add something like 
```c++
#if __cplusplus < 202002L
#pragma message("Warning: jaspBase requires c++20 or newer!")
#endif
```
to warn people that compilation may fail if they use an older version?

Let me know if that also works on your system!

boutinb:
The jaspBase must be also compiled by the JASP project. It is there only c++17. So we would get there a warning... Maybe we should also go to c++20 for JASP.  It seems to be supported by GCC, CLang and MSVC.

vandenman:
> It is there only c++17. So we would get there a warning...

Inside JASP we install jaspBase like any other R package, so it would still use C++20 if we specify that. I'm not sure if mixing C++ versions will cause problems.

JorisGoosen (approved):

PR #102, opened 2022-11-18
fix installing from renv

vandenman:
The construction

```r
if (!isModulePkgArchive(modulePkg) && correctlyInstalled)
  writeMd5Sums(modulePkg, moduleLibrary)
```
is also used inside `installJaspModuleFromDescription`. I'm probably the only one that noticed, but `isPkgArchive` never exists so `installJaspModuleFromRenv` was pretty broken.


R/moduleInstall.R
```diff
@@ -38,12 +38,12 @@ installJaspModule <- function(modulePkg, libPathsToUse, moduleLibrary, repos, on
   # if that isnt the case the (module) installation fails obscurely with an error like "Error in if (eval(cond, envir = environment(dot))) return(eval(expr, envir = environment(dot))): the condition has length > 1"
   cmdConfigCC <- system2(c(renv:::R(),"CMD","config","CC"),stdout=TRUE,stderr=TRUE)
   if(length(cmdConfigCC) > 1)
-    stop(
-"R CMD config CC returns more than 1 line, this will break renv and thus your install. 
-Most likely you are on mac and you should run `xcode-select --install` in a terminal. 
+    (if (getOS() == "osx") stop else warning)(
```
JorisGoosen, line 41:
```suggestion
    stop(
```

JorisGoosen (approved):
if the unittest pass this can be merged

PR #108, opened 2022-12-05
Fix mistake

Kucharssim:
I made a mistake here https://github.com/jasp-stats/jaspBase/pull/106 because apparently I don't know the difference between TRUE and FALSE.


vandenman (approved):
haha I've made that mistake more than once :smile:

PR #109, opened 2022-12-05
Add version argument to runWrappedAnalysis

boutinb:


R/common.R
```diff
@@ -1003,20 +1003,22 @@ registerData <- function(data) {
...
-      options <- checkAnalysisOptions(analysisName, options)
-      return(jaspTools::runAnalysis(analysisName, data, options))
-   }
-}
+runWrappedAnalysis <- function(analysisName, data, options, version) {
+  if (jaspResultsCalledFromJasp()) {
 
+    result <- list("options" = options, "analysis" = analysisName, "version" = version)
+    result <- jsonlite::toJSON(result, auto_unbox = TRUE, digits = NA, null="null", force = TRUE)
+    return(toString(result))
```
vandenman, line 1016: I don't think `toString` exists as a standalone function (it's a method on jaspObject)?

boutinb: > I don't think `toString` exists as a standalone function (it's a method on jaspObject)?
Hmm... I'm not sure to understand: first jaspObject is in jaspBase, so it should be anyway available. Second the `toString` function does not seem to come from jaspObject anyway. I can just call toString on a jsonlite object in a new session in R Studio without problem.

Kucharssim: There is a function `base::toString`:

> This is a helper function for `format` to produce a single character string describing an R object.

If you just want to convert the json object to a character, just use `as.character(result)`...

PR #110, opened 2022-12-06
remove calls to `.v` and `.unv` inside jaspBase

vandenman:
I see this one a bit too often:

image

so I figured, let's just remove the `.v` and `.unv` calls.
I did not try to clean up any of the code, so there are some things that may look a little bit odd now. For example, `targetB64 <- target`. Let's leave cleaning that up for another time.

### DIFF
--- a/R/common.R
+++ b/R/common.R
@@ -357,7 +357,7 @@ isTryError <- function(obj){
   if (is.null(new.df))
     return (data.frame())
 
-  names(new.df) <- .v(namez)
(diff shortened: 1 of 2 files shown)


Kucharssim (approved):

vandenman:
I'm expecting jaspMixedModels to fail, but the other modules should pass

Kucharssim:
I think it should be fine to merge!

PR #111, opened 2022-12-06
add default values for .numDecimals etc when not running in JASP

JorisGoosen:
For https://github.com/jasp-stats/INTERNAL-jasp/issues/2172

### DIFF
--- a/R/common.R
+++ b/R/common.R
@@ -92,6 +92,11 @@ runJaspResults <- function(name, title, dataKey, options, stateKey, functionCall
     oldwd                 <- getwd()
     setwd(location$root)
     withr::defer(setwd(oldwd))
+  } else {
+    .numDecimals        <- 3
+    .fixedDecimals      <- FALSE
+    .normalizedNotation <- TRUE
+    .exactPValues       <- FALSE
   }
 
   analysis    <- eval(parse(text=functionCall))
--- a/R/common.R
+++ b/R/common.R
@@ -92,7 +92,9 @@ runJaspResults <- function(name, title, dataKey, options, stateKey, functionCall
     oldwd                 <- getwd()
     setwd(location$root)
     withr::defer(setwd(oldwd))
-  } else {
+  } 
+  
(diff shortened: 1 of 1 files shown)


vandenman:
I mean this might work but it isn't a really nice solution.

What we need to have on the R side have some object (or function), say `jaspOptions`, that analyses can ask for global preferences. Then we don't have to pollute the global environment so much.

Whenever analyses then call `jaspOptions("numDecimals")` we retrieve the value for that preference (which can be cached of course).

shall I try to add that after this is merged?

JorisGoosen:
> I mean this might work but it isn't a really nice solution.
> 
> What we need to have on the R side have some object (or function), say `jaspOptions`, that analyses can ask for global preferences. Then we don't have to pollute the global environment so much.
> 
> Whenever analyses then call `jaspOptions("numDecimals")` we retrieve the value for that preference (which can be cached of course).
> 
> shall I try to add that after this is merged?

You are right this isn't pretty but it is also what we've been doing for a while now. The polluting that is.

The jaspOptions or whatever seems like a great solution but I think it'd be better if we made an issue for that and postpone it until at least the next version. And this can be done together with the splitting of of QMLComponents etc.

There we need to make all sorts of settings available in any case.

vandenman:
ok, I created an issue so we can let the topic rest for now.

vandenman:
I'm not 100% sure if this will work correctly, although I think it will (you're basically assuming that the lookup mechanism of R finds these values). Either way, this won't break existing analyses that don't use it so it's pretty harmless. If you'd want to be more safe you could put it in a `if (jaspResultsCalledFromJasp()) { ... }` instead.

JorisGoosen:
> you're basically assuming that the lookup mechanism of R finds these values

Indeed I am, as the function call is right below it and scoping in R simply looks in the environment up when it can't find something right?

I agree it is ugly but the new issue you made will entail cleaning this up anyway.

vandenman (approved):

PR #113, opened 2022-12-13
Fix save as pdf & save as eps, do not check for `capabilities("X11")`

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2136

Reverts https://github.com/jasp-stats/jaspBase/commit/6cc4b333fedf154ae6d3e54bfb885082351ecf98

@boutinb you added this change, but X11 is not a necessity for using cairo. In fact on Windows 10 I have that `capabilities("X11")` returns `FALSE` but all the cairo devices work. So for now I think we should revert the changes and try to find another solution to the other issue.

Also, note that this does not fix the problem 100%, whenever `device <- grDevices::pdf` (so `type != "cairo"`) the same error will reappear. I wonder if we can just assume/ ensure that we can always use cairo devices because they work better and as far as I cairo is shipped with R.


JorisGoosen:
While saving as pdf partly works now, in that it doesnt give the error it now opens a quartz window and doesnt save a file at all: 
image

When I try to save as eps I get an opened quartz window but completely empty with the error:
image

JorisGoosen:
it does work fine on windows btw.

Ill see if I can get cairo to behave on macos

R/common.R
```diff
@@ -189,10 +189,10 @@ registerFonts <- function() {
     # systemfonts::register_font(fontName, normalizePath(fontFile))
     # jaspGraphs::setGraphOption("family", fontName)
 
-    if (exists(".resultsFont"))
-      jaspGraphs::setGraphOption("family", .resultsFont)
+    if (exists(".resultFont"))
+      jaspGraphs::setGraphOption("family", .resultFont)
     else
-      warning("registerFonts was called but resultsFont does not exist!")
+      warning("registerFonts was called but resultFont does not exist!")
```
vandenman, line 195: Great that you updated the translation everywhere! However, this is exactly the kind of warning that needs `domain = NA` because it shouldn't be translated... That only makes our job harder when debugging and this warning isn't shown to users anyway.

shun2wang: Don't worry,  we dont needs `domain = NA` anymore :-) now will also remove the extra translation from the po.

boutinb:
This works on my Mac M1. I have tried all sorts of images.

R/common.R
JorisGoosen: I just went replace-all ^^

PR #118, opened 2023-01-11
Do not use tabs in Makevars files when specifying C++20

vandenman:
`CXX20STD` should be set by `CXX_STD` automatically. But if we use tabs then `CXX_STD` is ignored entirely, at least on Linux.

@boutinb can you check that this doesn't break anything on mac?


JorisGoosen (approved):

JorisGoosen:
works on macos

PR #120, opened 2023-02-14
make jaspResultsR inherit from jaspContainerR

vandenman:
I had to move the definition of `jaspResultsR` below that of `jaspContainerR` because otherwise I can't inherit from it. The second commit shows all the deduplicated code.


JorisGoosen (approved):
Looks good!

PR #125, opened 2023-06-01
fix .xxxSeedJASP

FBartos:
the old code was wrong and with new version of R fails on coercion:

```
Error in is.list(options) && c("setSeed", "seed") %in% names(options) : 
  'length = 2' in coercion to 'logical(1)'
```


FBartos:
how could this cause conflict in the line I change?

vandenman:
> how could this cause conflict in the line I change?

because you're 157 commits behind master :)

image

If you rebase this probably is fixed. Alternatively, use the web editor.

vandenman (approved):

PR #127, opened 2023-06-01
add mathExpression

Kucharssim:
This is used to avoid R programmers using the LaTeX delimiters directly. That way we won't collide with R stuff like `\()` and so on in the code, or change the delimiters easily if we have any reason to do so in the future.


JorisGoosen (approved):

Kucharssim:
It does not seem like the failing tests are caused by this PR...

PR #128, opened 2023-06-02
Fix import statements and is.jaspRhs

vandenman:
I wondered why the unit tests failed and figured, let's clean up some of the notes. It turns out the one error had nothing to do with the imports, but I had already fixed them so let's also push them.

The main issue was that in this function
```r
is.jaspRhs <- function(x) {
  if(inherits(x, "list")) {
    all(vapply(x, is.jaspRhs, logical(1)))
  } else {
    inherits(x, "jaspFormulaRhs")
  }
}
```
`inherits(x, "list")` will also return TRUE for objects of class `jaspFormulaRhs`. So we ended up `vapply`ing over the elements of a `jaspFormulaRhs` and those are no longer lists nor `jaspFormulaRhs`s. This caused an example to fail and that's where the recent red checkmarks come from. So essentially, I reversed the order of the list test/


Kucharssim (approved):
Ah yeah, thanks for fixing this!

PR #134, opened 2023-09-07
Use testthat

Kucharssim:
While working on https://github.com/jasp-stats/INTERNAL-jasp/issues/2332, I need to write tests because otherwise I drown trying to make sure I cover all functionality and corner cases.

This PR only enables `testthat` for this repo and adds just a few tests for already existing functionality. The tests are run automatically as part of the R-CMD-CHECK action.

### DIFF
--- a/.github/workflows/downstream-tests-on-comment.yml
+++ b/.github/workflows/downstream-tests-on-comment.yml
@@ -88,7 +88,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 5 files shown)


vandenman (approved):
looks good to me! Perhaps it's also a good idea to update the R version in the workflow file?

Kucharssim:
> Perhaps it's also a good idea to update the R version in the workflow file?

Ok, bumped it up to 4.3.1

PR #136, opened 2023-09-11
Column type conversions

Kucharssim:
fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2332

Not sure if this is really the best approach, but it does resolve the issue. I still need to check if really everything works as expected, I welcome any input.


R/readDataSet.R
```diff
@@ -0,0 +1,66 @@
+#' @rdname column-types
+#' @export
+setDataSet <- function(dataset) {
+  .internal[["dataset"]] <- as.data.frame(lapply(dataset, r2jasp))
```
vandenman, line 4: perhaps we should add convenience methods to convert entire data.frames?

```r
r2jasp.data.frame <- function(x) {
  as.data.frame(lapply(x, r2jasp))
}

jasp2r.data.frame <- function(x) {
  as.data.frame(lapply(x, jasp2r))
}
```

vandenman (changes requested):
This looks really good! 

However, this is a pretty aggressive change. I wonder whether we can create an opt-in way for adding this to a module, like how we once had `jaspResults: true` in the QML files. Because I can imagine that many hidden issues appear when this is suddenly some R package receives a `jaspScale` instead of a `numeric` and that we need to sprinkle some `jasp2r` calls around.

vandenman:
let's see if the downstream tests action is actually still working :sweat_smile:

Kucharssim:
> Because I can imagine that many hidden issues appear when this is suddenly some R package receives a jaspScale instead of a numeric and that we need to sprinkle some jasp2r calls around.

I am not sure if I am overlooking a case that you have specifically in mind. But in general the code inside of modules will still run on base R types as the data is obtained using `readDataSetToEnd` which should return either `numeric`, `ordinal`, or `factor`. This whole jaspType shenanigans is used only to simulate `readDataSetToEnd`'s behavior when run from R.

Unless of course we want to fully switch to JASP column types inside of analyses but then indeed we would need to implement this gradually because the current module code is not written for it.

vandenman:
> I am not sure if I am overlooking a case that you have specifically in mind.

Well, unless the GitHub action is doing something wrong, this PR breaks all unit tests from all modules. So my use case is, do we need to change something in the modules (or jaspTools) to make this PR not break everything? If it needs to be done in each module, I would imagine we prefer to do that incrementally.

Kucharssim:
I'll check what's wrong in the tests, the PR should definitely not break this.

Kucharssim:
Ah, of course. I forgot that `jaspTools` tries to load the dataset through implementing its own `readDataSetToEnd` that is found by `jaspBase` using `getAnywhere()` - a connection that I broke with this PR.

So this PR will need to be accompanied by a little change to `jaspTools` so that the unit tests are able to load the correct dataset.

Anyway, I see that some of the downstream tests did not even make it to run the tests because `stringi` could not be installed/loaded. I think `stringi` is suggested by `pillar` which is used for some of the pretty printing of the jaspTypes in R console, but I guess I can remove that. What do you think @vandenman?

shun2wang:
I think I once fixed this error with `stringi` see https://github.com/jasp-stats/INTERNAL-jasp/issues/2192, the libicu library which is dependence of R packages such as `igraph` on Linux I guess.

so I made: https://github.com/jasp-stats/jasp-actions/pull/30

Kucharssim:
So in combination with https://github.com/jasp-stats/jaspTools/pull/38 the common modules now pass their tests on my computer. I will test the rest as well just to make sure that all is really ok.

vandenman:
:man_facepalming: this is not using the new jaspTools so of course all the tests fail...

Kucharssim:
> this is not using the new jaspTools so of course all the tests fail...

yes that's why I was running the tests manually on my computer. Can we specify a specific branch/commit of `jaspTools` in the downstream tests, or do you want to just merge the jaspTools PR first and then run the tests here? Note that once we merge the jaspTools PR, tests will be broken without the changed to jaspBase

Kucharssim:
@vandenman @JorisGoosen this is basically outdated, right? Should I close this so that it doesn't hang around here forever?

vandenman:
Nah leave it open, I'd still like to implement this someday. Perhaps if Joris adds another option for QML to enable this we can look at doing this more gradually though?

PR #140, opened 2023-11-15
add some rowwise functions for jasp-stats/jasp-issues#516

JorisGoosen:
Are these ok @Kucharssim ?


Kucharssim (commented):
This will probably work for the basic use case, I just have some observations about the design of this:

1. You could potentially use `rowSums()` and `rowMeans()` that are already available in base R.
2. How do we want it to behave if someone puts a nominalText vector in, e.g., `rowMean`? Do we throw an error? Do we silently coerce it to numeric? Do we drop it from the computation?
3. What if someone wants to take a product of multiple columns? Or compute a geometric mean? Would we manually add `rowProduct`, `rowGeometricMean`, etc.? Perhaps we could instead provide something more generic that would behave like `apply`? Like we could for instance have a function with signature `rowWise(function, ...)` that would row-wise apply `function` to columns in `...`, or something like that... Obviously we probably need to make sure that it would work for different functions, etc., and I am not sure if it's not reinventing `apply`, but it feels like something could be done here. Any opinions @vandenman?

JorisGoosen:
Well, those are all good points and I invite you to make the required changes to the branch.
I figured I am not the best R programmer but that this would already be much better than not having simple functions like this in JASP.

Any improvements you can make are welcome

JorisGoosen:
> How do we want it to behave if someone puts a nominalText vector in, e.g., rowMean? Do we throw an error? Do we silently coerce it to numeric? Do we drop it from the computation?

I figured this could also be thought of later, in use.
Also it wouldnt make much sense to put nominalText in any of these anyway right?

JorisGoosen:
> What if someone wants to take a product of multiple columns? Or compute a geometric mean? Would we manually add rowProduct, rowGeometricMean, etc.? Perhaps we could instead provide something more generic that would behave like apply? Like we could for instance have a function with signature rowWise(function, ...) that would row-wise apply function to columns in ..., or something like that... Obviously we probably need to make sure that it would work for different functions, etc., and I am not sure if it's not reinventing apply, but it feels like something could be done here. Any opinions @vandenman?

Well this would just be apply no?

the point of these functions is to be a middleground between the drag&drop constructor and R.
The only reason Im not adding it to the column constructor is because the code is my first attempt at QML and a bit of a mess... So this is the next best thing

Kucharssim:
> I figured I am not the best R programmer but that this would already be much better than not having simple functions like this in JASP.

For sure, I don't have a problem with going with this implementations, I just thought we could think about a possible improvements later.

> I figured this could also be thought of later, in use.

Sure

> Also it wouldnt make much sense to put nominalText in any of these anyway right?

Yeah, but that doesn't stop users from trying to do that 😛

> Well this would just be apply no?
> the point of these functions is to be a middleground between the drag&drop constructor and R.
The only reason Im not adding it to the column constructor is because the code is my first attempt at QML and a bit of a mess... So this is the next best thing

Well, kind of. If we just exposed `apply` we would also have to expose `data.frame`/`cbind` or something, so that you can do:

```
apply(cbind(contNormal, contGamma), 1, mean)
```

With `rowWise`, you would just do:

```
rowWise(mean, contNormal, contGamma)
```

Which would be imo the middleground between drag&drop and full on R.
 
But it would definitely require some more thought on the design of this, so probably good to discuss with @vandenman as well. In the meantime, what you created will definitely be better than nothing.

NAMESPACE
```diff
@@ -122,5 +122,21 @@ export(unv)
 export(v)
 export(weibullDist)
 export(zScores)
+export(rowMean)
```
vandenman, line 125: please regenerate this with roxygen2 instead of editing the NAMESPACE file manually.

R/friendlyConstructorFunctions.R
```diff
@@ -149,3 +149,84 @@ hasSubstring <- function(string, substring) {
...
+
+#' @export
+rowSum <- function(...) {
+  apply(data.frame(...), 1, sum, na.rm=FALSE)
+}
+
+#' @export
+rowSumNaRm <- function(...) {
+  apply(data.frame(...), 1, sum, na.rm=TRUE)
+}
```
vandenman, line 171: Here is a minor optimization
```suggestion
rowMean <- function(...) {
  rowMeans(matrix(c(...), ncol = ...length()), na.rm=FALSE)
}

#' @export
rowMeanNaRm <- function(...) {
  rowMeans(matrix(c(...), ncol = ...length()), na.rm=TRUE)
}

#' @export
rowSum <- function(...) {
  rowSums(matrix(c(...), ncol = ...length()), na.rm=FALSE)
}

#' @export
rowSumNaRm <- function(...) {
  rowSums(matrix(c(...), ncol = ...length()), na.rm=TRUE)
}
```
`rowMeans` and `rowSums` are optimized versions for the very common `apply(x, 1, mean/sum)`. The R package `matrixStats` has a lot more of these (e.g., `rowQuantiles`). We probably use that package anyway so it's worthwhile to consider here.

R/friendlyConstructorFunctions.R
```diff
@@ -149,3 +149,84 @@ hasSubstring <- function(string, substring) {
...
+
+#' @export
+rowSD <- function(...) {
+  apply(data.frame(...), 1, sd, na.rm=FALSE)
```
vandenman, line 175:
```suggestion
  apply(data.frame(...), 1, stats::sd, na.rm=FALSE)
```
the same goes for everything until `min`.

vandenman (changes requested):

JorisGoosen:
> In the long run, we should probably rethink the entire setup for the compute column situation. I can also imagine that we create our own small minilanguage like Excel/ Google Docs instead of a drag and drop. The next thing I imagine people ask for is something like compute this after grouping on that variable and then we get very close to dplyr.

Yes and we should probably rewrite it entirely, same for filter.

JorisGoosen:
Or just make R syntax completion available inside the editors and some good tutorials

vandenman:
I changed whatever I thought was necessary. The construction `matrix(c(...), ncol = ...length())` is a bit more eager to convert everything to numeric than `data.frame`. For example:

``` r
a <- factor(1:4)
b <- rnorm(4)

# what matrix(c(...), ncol = ...length()) does
cbind(a, b)
#>      a          b
#> [1,] 1 -0.4974208
#> [2,] 2 -0.7989079
#> [3,] 3  0.1453013
#> [4,] 4 -0.5237008

foo_1 <- function(..., fun = mean) {
  apply(data.frame(...), 1, fun)
}

foo_2 <- function(..., fun = mean) {
  apply(matrix(c(...), ncol = ...length()), 1, fun)
}

foo_1(a, b)
#> Warning in mean.default(newX[, i], ...): argument is not numeric or logical:
#> returning NA

#> Warning in mean.default(newX[, i], ...): argument is not numeric or logical:
#> returning NA

#> Warning in mean.default(newX[, i], ...): argument is not numeric or logical:
#> returning NA

#> Warning in mean.default(newX[, i], ...): argument is not numeric or logical:
#> returning NA
#> [1] NA NA NA NA
foo_2(a, b)
#> [1] 0.2512896 0.6005460 1.5726507 1.7381496

foo_1(a, b, fun = sum)
#> Error in FUN(newX[, i], ...): invalid 'type' (character) of argument
foo_2(a, b, fun = sum)
#> [1] 0.5025792 1.2010921 3.1453013 3.4762992

foo_1(a, b, fun = sd)
#> [1] 1.058836 1.979127 2.018577 3.198740
foo_2(a, b, fun = sd)
#> [1] 1.058836 1.979127 2.018577 3.198740
```

In addition, I removed `rowCovariance` and rowCorrelation`. You'd need two vectors for those functions, and I don't see how that was supposed to work.

PR #147, opened 2024-04-29
Fix building failure with `-Werror=format-security`

sukanka:
Archlinux enables `-Werror=format-security` by default, which causes jaspBase failed to be built. This pr fix it.
Here is a [relevant log](https://build.bioarchlinux.org/api/pkg/r-jaspbase/log/1714092710) (at the bottom of the page).


shun2wang:
@JorisGoosen pushed a fix for it so I think no need for it now.

but thanks for your PR anyway.

JorisGoosen:
@sukanka thanks for the PR!

I had simply not noticed it in my flood of github emails...

PR #151, opened 2024-06-18
Add `readDataSetByVariableTypes`

vandenman:
@JohnnyDoorn, @FBartos this PR adds a new function `readDataSetByVariableTypes` for reading the data while respecting user-selected data types.

Right now, every variable passed from QML adds an element to the options list with the name {name}.type. So "variables" -> "variables.types", "split" -> "split.types", etc. Conveniently (for me) this is also done for variables with a fixed type that the user cannot change.

The currently intended usage is
```r
dataset <- readDataSetByVariableTypes(options, c("variables", "split", ...))
```
If a user only entered something in "variables", but "split" is empty, then this is automatically ignored.

The argument `exclude.na.listwise` is forwarded to `.readDataSetToEnd`.

It is not possible to read the same variable in more than once (if this happens an error is thrown).

This will obviously require some careful thinking (and testing), so any thoughts and suggestions are welcome!


JohnnyDoorn:
Awesome, so this avoids needing to tweak `.readDataSetToEnd` everywhere and specifying the column types manually? (just for my own understanding)

vandenman:
Yes, exactly. But if the qml options only allow for one type anyway (e.g., in linear regression), then you don't need to use this and can call `.readDataSetToEnd` directly because you know which options are read as which type. But if you have right now `.readDataSetToEnd(columns=...)`, like in descriptives and raincloudplots, then you need to look at the associated types, and this function takes care of that for you.

JohnnyDoorn (approved):

JohnnyDoorn:
On the surface this looks good, although I cannot test it myself because i'm still in the process of setting up building JASP

JorisGoosen (approved):

PR #153, opened 2024-06-24
Automatically call optionContainsValue for elements <>.types

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2594 in general for all analyses.

I would revert https://github.com/jasp-stats/jaspDescriptives/commit/7dff69664279428f002e47e3106156701a3125ac, because that causes all plots to be needlessly recomputed whenever a user adds a variable.

I'm not sure if there ever is a need to _not_ do this.


vandenman:
actually, I think this needs to be done differently, right now it'll just check if any type matches, rather than the type of the specific variable.

vandenman:
superseded by the new data reading stuff.

PR #156, opened 2024-09-18
generate version info for jaspColumnEncoder

vandenman:
Dynamically create `jaspBase:::jaspColumnEncoderInfo()`. This will print the output of `git log -1 --oneline` if obtained from a git module. If `jaspColumnEncoder` was obtained differently, it'll print how it was obtained instead.

### DIFF
--- a/configure
+++ b/configure
@@ -21,6 +21,7 @@ if [ "$UNAME" = "SunOS" ]; then
 fi
 
 DOWNLOAD_SUCCESS=1
+EXTRA_INFO="Version info is only available when git is available!"
 
 if [ "${INCLUDE_DIR}" ]; then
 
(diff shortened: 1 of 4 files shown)


configure
```diff
@@ -29,6 +30,16 @@ if [ "${INCLUDE_DIR}" ]; then
 	PKG_CXXFLAGS="-I\"${INCLUDE_DIR}\""
 	JASPCOLUMNENCODER_DIR="${INCLUDE_DIR}"
 
+	if git --version 2>&1 >/dev/null; then
+
+		if [ -d "${JASPCOLUMNENCODER_DIR}/.git" ]; then
+			EXTRA_INFO="include_dir, git_repo at:  + $(git --git-dir="${JASPCOLUMNENCODER_DIR}/.git" --work-tree="${JASPCOLUMNENCODER_DIR}" log -1 --oneline)"
+		else
+			EXTRA_INFO='include_dir that is not a git repo.'
```
JorisGoosen, line 38: If its not from git we are copying it from somewhere else right?
Maybe print the include path as well for the non-git case?

configure.win
```diff
@@ -53,6 +67,7 @@ else
 
 			curl --silent --location https://api.github.com/repos/jasp-stats/jaspColumnEncoder/tarball | tar xz --strip=1 --directory inst/include/jaspColumnEncoder
 			DOWNLOAD_SUCCESS=$?
+			EXTRA_INFO="curl was used to download jaspColumnEncoder, no version info is available."
```
JorisGoosen, line 70: Ah, I guess these do something like that

JorisGoosen (approved):

PR #165, opened 2025-03-05
cleanup dependencies

vandenman:
This should be tested carefully.

```r
master <- pkgdepends::pkg_deps$new("jasp-stats/jaspBase")
thisPr <- pkgdepends::pkg_deps$new("vandenman/jaspBase@cleanDependencies")

master$solve()
thisPr$solve()

master
thisPr

> master
<pkg_dependencies>
+ refs:
  - jasp-stats/jaspBase
+ has resolution (+113 dependencies)
+ has solution
(use `$get_resolution()` to see resolution results)
(use `$show_solution()` to see the dependencies
(use `$get_solution()` to see the full solution results)
(use `$draw()` to draw the dependency tree)
> thisPr
<pkg_dependencies>
+ refs:
  - vandenman/jaspBase@cleanDependencies
+ has resolution (+62 dependencies)
+ has solution
(use `$get_resolution()` to see resolution results)
(use `$show_solution()` to see the dependencies
(use `$get_solution()` to see the full solution results)
(use `$draw()` to draw the dependency tree)
```
So going from 113 base dependencies to 62 is a pretty nice win. Most of the gain is in qgraph.
It's also possible to clean this up even more, but that will take more time (and rewriting some code).

PR #166, opened 2025-03-05
Allow the `plot` object passed to `jaspPlot` to be a try error

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/516

basically some sugar to transform
```r
plot <- try({ggplot stuff})
if (inherits(plot, "try-error"))
  jaspPlot$setError(.extractErrorMessage(plot))
else
  jaspPlot$plotObject <- plot
```
into
```r
jaspPlot$plotObject <- try({ggplot stuff})
```
This also works in the constructor `createJaspPlot(plot = try({...}))` because that calls `jaspPlotObj$plotObject <- plot` to set the plot.

### DIFF
--- a/R/zzzWrappers.R
+++ b/R/zzzWrappers.R
@@ -598,13 +598,12 @@ jaspPlotR <- R6::R6Class(
       if (!is.null(error))
         jaspPlotObj$setError(error)
 
-      if (!is.null(plot))
-        jaspPlotObj$plotObject <- plot
+      private$setPlotObject(plot, setNull = FALSE)
 
       if (!is.null(info))
         jaspPlotObj$info <- info
 
-      if(is.numeric(position))
+      if (is.numeric(position))
         jaspPlotObj$position = position
 
       super$dependOn(dependencies)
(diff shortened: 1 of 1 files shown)


vandenman:
tests should pass after https://github.com/jasp-stats/jaspBase/pull/165

vandenman:
On second thought, we could also put the try inside of jaspPlotR. Then we could write
```r
jaspPlot$plotObject <- create_ggplot()
```
and then the developer doesn't even need to write the try (this works because R is lazy). 

@FBartos, as the only other person that might ever use this, would that be too magical or would you prefer that?

FBartos:
> On second thought, we could also put the try inside of jaspPlotR. Then we could write
> 
> ```r
> jaspPlot$plotObject <- create_ggplot()
> ```
> 
> and then the developer doesn't even need to write the try (this works because R is lazy).
> 
> @FBartos, as the only other person that might ever use this, would that be too magical or would you prefer that?

I don't usually slot the plotting output into the jaspPlot$plotObject directly (since I want to inspect those during debugging, etc), but it might he helpful for some uses. Definitelly the first improvement, automatically setting errors simplifies the written code nicely.

I was wondering if we could have the same thing for tables and htmlOutputs for completeness. For a table, addRow or setData would turn it into a setError if any of the inputs is error (and any additional addRow calls would be ignored)

vandenman:
ok so now it works when doing `createJaspPlot(plot = foo())`, but not when doing 
```r
plot <- createJaspPlotObject()
plot$plotObject <- foo()
```
because `$` always evaluates `foo()` eagerly. This is probably intentional in R, but I'll nevertheless ask around. One way around this is to deprecate the active method `plot$plotObject` and instead always use an explicit method (e.g., `plot$setPlotObject(foo())`).

@FBartos I'm not sure if `plot$setPlotObject(foo())` is worth the effort, what do you think?

FBartos:
would be nice but not worth it IMO

PR #176, opened 2025-08-29
Add automatic transforms

Kucharssim:
Part of a fix for Gitea SGC#15 and prerequisite for Gitea SGC#6

@JTPetter: For the Box-Cox, I added an option to estimate lambda as described in the Minitab manual: https://support.minitab.com/en-us/minitab/help-and-how-to/quality-and-process-improvement/control-charts/how-to/box-cox-transformation/methods-and-formulas/methods-and-formulas/. But I don't have Minitab. Would it be possible to check whether this implementation indeed gives similar results (there will always probably be some small numerical differences though).


Kucharssim:
Thanks for checking!

Summary of the changes since your testing:
1. The minitab version of the Box-Cox is now only defined as x^lambda instead of the standard box-cox (x^lambda - 1)/lambda. I've also documented these details in the help files.
2. The minitab version can now do estimation of lambda for grouped data (using pooled sd), as well as "invidual data" (using the unbiased average moving range)

I've also added the Johnson transform (not to be confused with Yeo-Johnson: different methods!) that is also apparently provided by Minitab. The two implementations will probably differ though (Minitab uses different criterion for selecting the optimal parameters of the transform than the [original paper](https://www.tandfonline.com/doi/abs/10.1080/00224065.1998.11979832)). As far as I can see in the docs, this transform is only used within Minitab to tranform the data regardless of the data subgroups. The function also returns some additional info in the attributes, which will be useful for integrating this transform in the [process capability study](https://support.minitab.com/en-us/minitab/help-and-how-to/quality-and-process-improvement/capability-analysis/how-to/capability-analysis/normal-capability-analysis/methods-and-formulas/johnson-transformed-data/) analysis.

https://github.com/jasp-stats/jasp-desktop/pull/6000 needs to reflect the recent changes but this PR is ready for review again.

JTPetter (approved):

PR #204, opened 2026-05-14
Support source-module wrapped analysis replay

FBartos:
﻿## Summary

This PR provides the jaspBase runtime side of the jaspTools -> jaspSyntax bridge:

- lets `runWrappedAnalysis()` resolve analyses from explicit source module paths and QML files, not only installed module packages
- persists and retrieves standalone analysis state through the native callback-file contract instead of reaching into jaspTools internals
- carries the opaque jaspSyntax/SyntaxInterface column encoder context through result materialization
- decodes R-facing result copies returned by `$toRObject()`, including JASP-owned table names, footnotes, stored result-state figures, and ggplot labels/data/mappings/metadata
- keeps encoded names in the backend/runtime path while avoiding deep mutation of opaque analysis-owned model objects
- treats missing or failing native decoding as a bug for encoded tokens, rather than falling back to R-side maps
- avoids double-rendering function/base plots when writing images

## Why

Desktop/SyntaxInterface owns token replacement, jaspSyntax exposes that bridge, and jaspBase owns the boundary between backend analysis objects and R-facing result copies. This PR keeps that boundary explicit: JASP-owned display/result surfaces are decoded eagerly from the captured encoder context, while arbitrary downstream module internals stay opaque.

That fixes source-module replay and `toRObject()` decoding without adding downstream module dependencies to core packages or duplicating Desktop encoding rules in R.

## Related PRs

- jasp-desktop native encoder context: https://github.com/jasp-stats/jasp-desktop/pull/6249
- jaspSyntax encoder-context R API: https://github.com/jasp-stats/jaspSyntax/pull/8
- jaspTools replay orchestration: https://github.com/jasp-stats/jaspTools/pull/79
- jaspDescriptives replay-test blocker fix: https://github.com/jasp-stats/jaspDescriptives/pull/500

## Verification

- focused jaspBase tests passed: `test-result-object-decoding.R`, `test-runWrappedAnalysis.R`
- focused jaspSyntax tests passed against the rebuilt local Desktop bridge: `test-desktop-jasp-contract.R`, `test-dataset-helpers.R`
- focused jaspTools lifecycle tests passed: `test-jaspSyntax-lifecycle.R`
- R parse check passed
- `git diff --check`


R/common.R
```diff
@@ -1163,18 +1166,49 @@ storeDataSet <- function(dataset) {
...
     # in a structured way. In this way the Desktop can then set the options to the QML controls of the form, and this will run the analysis.
     # So here, just give back the parsed options.
-    return(toJSON(list("options" = options, "module" = moduleName, "analysis" = analysisName, "version" = version)))
+    response <- list(
+      "options"         = options,
+      "module"          = moduleName,
+      "analysis"        = analysisName,
+      "version"         = version,
+      "qmlFileName"     = qmlFileName,
+      "jaspBaseVersion" = as.character(utils::packageVersion("jaspBase")),
```
vandenman, line 1360: Was this already the jaspBase version, or is it the version from the jasp module?

R/common.R
```diff
@@ -1163,18 +1166,49 @@ storeDataSet <- function(dataset) {
   jaspSyntax::loadDataSet(dataset)
 }
 
+.wrappedAnalysisQmlFile <- function(moduleName, qmlFileName, modulePath = NULL, qmlFile = NULL) {
+  isNonEmptyString <- function(x) is.character(x) && length(x) == 1 && !is.na(x) && nzchar(x)
```
vandenman: I'm wondering if we shouldn't use rlang for many of these predicates and fs for file system operations. We import those packages no matter what and they might clean up a lot of these verbose helpers and file lookup functions.

R/common.R
```diff
@@ -659,19 +660,20 @@ jaspResultsStrings <- function() {
 
 }
 
+.stateFilePath <- function(location) {
+  if (is.list(location) && !is.null(location$root) && !is.null(location$relativePath))
```
vandenman: This feels very verbose. Perhaps add a comment with why there are so many checks? Are other possible values allowed? If yes, shouldn't this be streamlined elsewhere?

vandenman (changes requested):

R/common.R
FBartos: Agreed. I incorporated this in f51c8d7: `fs` and `rlang` are now declared imports, the string/list predicates use `rlang`, and the QML/state path handling uses `fs` for path construction, existence checks, normalization, and directory creation. The calls are namespace-qualified so the dependency use is explicit.

R/common.R
FBartos: Good catch. The existing `version` field is the generated module wrapper version, not the jaspBase version. I kept that field unchanged for compatibility and added `jaspBaseVersion` as the separate runtime-contract version. I also added an inline comment in f51c8d7 so this distinction is explicit at the call site.

R/common.R
FBartos: Agreed. I tightened this in f51c8d7. The helper now validates the callback shape and errors clearly for unsupported values. The intended Desktop/native contract is `list(root, relativePath)`, while a standalone callback may provide only `relativePath`, which is interpreted relative to the current working directory. I also added a focused test for unsupported callback shapes so this contract is covered.

R/writeImage.R
```diff
@@ -296,6 +297,39 @@ decodeplot.function <- function(x, ...) {
...
+
+  state
+}
+
+.decodeJaspPlotObject <- function(plot) {
+  tryCatch(
+    decodeplot(plot, returnGrob = FALSE),
+    error = function(e) plot
+  )
+}
```
vandenman: Two issues.

1. If a user loads a dataset, then does the analysis, it works. If they load another dataset afterward, and show then print/ save the plot from the previous analysis, this no longer works, no? We need some hook to know which encoding/ decoding object should be used for which plot object.
2. If we do save or saveRDS on the output from jasp, then restart the r session and then try to replay the plot, all the information about encoding/ decoding is no longer available because this lives on the C++ side of jaspSyntax. We need a way around this. This could be done by immediately storing the decodedplot and not doing it on demand.

FBartos:
Implemented the eager-decoding follow-up from the review/spec discussion.

What changed:
- Added an internal per-analysis decode context in `jaspBase` that captures the column mapping and factor-level mapping while the native/jaspSyntax state is still live.
- Materialize decoded state before saving: `finishJaspResults()` now decodes `figures` and `other` state, and `.saveState()` defensively enforces the same contract.
- Kept plot objects editable: ggplot/jaspGraphs/qgraph objects are stored as decoded editable objects, not rendered gTrees; base/function plots still materialize through recorded/grob behavior as before.
- Updated the C++ plot handoff so the decoded object returned by `writeImageJaspResults()` is written back into plot state for all plot types, not only function plots.
- Routed edit options and plotly conversion through the decoded object, so PNG/edit metadata/interactive JSON have the same labels.
- `decodeJaspResultState()` is no longer exported; the state repair/materialization helper is internal now.
- Extended R-facing object decoding to names, attributes, data frames, character vectors, factors, and numeric factor tokens using the captured context.

Companion jaspTools follow-up:
- Pushed jasp-stats/jaspTools#82 so the fast test plot writer also asks jaspBase for the decoded plot object instead of storing the raw encoded object.
- Fixed the viewer helper to create the html output directory before copying assets; this came up when replaying the MixedModels example with `view = TRUE`.

Verification:
- `R CMD INSTALL .` for jaspBase passes.
- `R CMD check --no-manual --no-vignettes .` completes and runs testthat. It still reports the existing source-tree/package-metadata warnings/notes, including the already-known undeclared optional `jaspSyntax` usage, but no test failures.
- Focused jaspBase tests pass: `test-result-object-decoding.R`, `test-runWrappedAnalysis.R`.
- Focused jaspTools tests pass: `test-runAnalysis-fast-test-plots.R`, `test-view.R`.
- Replayed `MixedModelsGLMM` from `Larks and Owls.jasp`; it returns results successfully and the returned state has no `JaspColumn_...`/`jaspColumn...` tokens.

FBartos:
Follow-up from the final desiderata audit: I found one remaining provenance gap in the live R6 wrapper path. 	oRObject() called the C++ 	oRObject() materializer first, and that C++ code can consult the currently active global decoder before the R-side eager decoder sees the object. That meant a live wrapper could still be decoded against the wrong dataset after an analysis/dataset switch, even though saved state was already eager-decoded.

Fixed in 17d2533:
- 
unJaspResults() stores the per-analysis decode context on the result wrapper and propagates it to child wrappers.
- R-facing 	oRObject() materialization now temporarily installs a context-backed internal decoder while C++ materializes the object, then applies the recursive R decoder with the same context. This keeps C++ behavior intact but prevents late active-dataset provenance leaks.
- saveImage(), editImage(), and 
ewriteImages() now treat loaded plot state as stored result state rather than consulting a live analysis decoder. If encoded legacy state is encountered without context, it warns and leaves it unchanged instead of silently decoding against the wrong dataset.
- Added a regression test where the active decoder intentionally maps the same token to a wrong dataset name; the R6 wrapper still returns the original analysis names and factor labels.

Verification:
- 	est-result-object-decoding.R passes against the installed temp-library jaspBase.
- 	est-runWrappedAnalysis.R passes.
- jaspTools focused fast-plot/view tests pass with the updated jaspBase source.
- MixedModels MixedModelsGLMM replay from Larks and Owls.jasp passes through jaspTools::runAnalysis() with no encoded state tokens.
- R CMD check --no-manual --no-vignettes exits 0; it still reports the existing source-tree warnings/notes unrelated to this change.

FBartos:
@vandenman this is ready for your review. The final desiderata pass is pushed, including the decode-context provenance follow-up for live result materialization and stored plot replay/export.

FBartos:
Follow-up after the direct `jaspMixedModels::MixedModelsLMM()` cake repro: this did point to a structural boundary issue in the decoder, not just an `lme4` edge case.

I pushed `0a64ba4` with three related changes:

- Decode only JASP-owned / display-facing result surfaces. `.decodeJaspRObject()` now preserves opaque S3/S4/call/name objects instead of recursively rewriting arbitrary package-owned internals. This avoids mutating `lme4` internals such as `lmerResp`.
- Treat JASP `mixed` table cells by structure, not by the bare class name. The repro also exposed that `afex::mixed()` returns model objects with class `mixed`, so using `inherits(x, mixed)` as provenance was wrong. The decoder now recognizes only JASP mixed vectors/cells and leaves foreign `mixed` model objects intact.
- Keep `state$other` opaque. That field contains `jaspState` payloads restored into later runs, so it is analysis-owned state. Plot replay/export/edit state under `state$figures` is still eagerly decoded.

I also moved the per-analysis decode-context capture until after dataset preload, because that is when `jaspSyntax` exposes the requested encoded column names. Without that, `toRObject()` had a context object but not the `JaspColumn_* -> original name` mapping.

Verification after the change:

- `pkgload::load_all('.', quiet = TRUE); testthat::test_file('tests/testthat/test-result-object-decoding.R')` passes.
- Installed `jaspBase` into `C:/JASP-Packages/_verify-lib-4.6` and reran the cake `MixedModelsLMM` repro. It returns `ANOVA Summary, Plot`, and scanning the R-facing `toRObject()` output finds no visible `JaspColumn_*` tokens.
- `R.exe CMD check --no-manual --no-vignettes .` completes with the existing source-tree warnings/notes; tests pass.

R/resultDecoding.R
```diff
@@ -83,14 +83,26 @@
...
-      error = function(e) character()
-    )
     requestedDataset <- tryCatch(
       getExportedValue("jaspSyntax", "readRequestedDataset")(decode = FALSE, normalize = FALSE),
       error = function(e) NULL
     )
+    defaultMapping <- tryCatch(
+      getExportedValue("jaspSyntax", "columnMapping")(strict = FALSE),
+      error = function(e) character()
+    )
```
vandenman: This should at least throw a warning, no? Same for the `requestedMapping` below?

R/common.R
```diff
@@ -820,7 +844,11 @@ saveImage <- function(plotName, format, height, width)
   state           <- .retrieveState()     # Retrieve plot object from state
   plt             <- state[["figures"]][[plotName]][["obj"]]
 
-  plt             <- decodeplot(plt);
+  plt             <- .decodeJaspPlotObject(
+    plt,
+    returnGrob = FALSE,
+    decodeContext = .jaspDecodeContext(source = "stored-result-state")
+  )
```
vandenman, line 851: In saveImage we should always use the returnGrob path because otherwise we might show encoded column names in plots within jasp.

R/common.R
```diff
@@ -1026,7 +1054,11 @@ rewriteImages <- function(name, ppi, imageBackground) {
 
       jaspPlotCPP$editing <- TRUE
 
-      plot <- jaspPlotCPP$plotObject
+      plot <- .decodeJaspPlotObject(
+        jaspPlotCPP$plotObject,
+        returnGrob = FALSE,
+        decodeContext = .jaspDecodeContext(source = "stored-result-state")
+      )
```
vandenman, line 1061: revert this? when reassigned to $plotobject this should do the de/ encoding then within writeImage?

R/common.R
```diff
@@ -1077,7 +1109,11 @@ editImage <- function(name, optionsJson) {
     jaspPlotCPP$editing <- TRUE
     on.exit({jaspPlotCPP$editing <- FALSE}) # this should not persist!
 
-    plot <- jaspPlotCPP$plotObject
+    plot <- .decodeJaspPlotObject(
+      jaspPlotCPP$plotObject,
+      returnGrob = FALSE,
+      decodeContext = .jaspDecodeContext(source = "stored-result-state")
+    )
```
vandenman, line 1116: revert this as well, see prior comments.

R/common.R
```diff
@@ -1163,27 +1201,194 @@ storeDataSet <- function(dataset) {
...
+    }, add = TRUE)
+
+    sink(outputConnection, type = "output")
+  }
+
+  if (showAnalysis)
+    return(.runWrappedAnalysisWithDecodedConditions(expr))
+
+  suppressWarnings(suppressMessages(expr))
+}
```
vandenman, line 1343: this works, but why not adjust https://github.com/jasp-stats/jasp-desktop/blob/22c167b122776652968722eaf7c09debb24e3203/Common/log.cpp#L18 so that we can disable or enable this in a static jasp-build from jaspSyntax?

R/writeImage.R
```diff
@@ -79,7 +79,8 @@ writeImageJaspResults <- function(plot, width = 320, height = 320, obj = TRUE, r
   width  <- width  * (ppi / 96)
   height <- height * (ppi / 96)
 
-  plot2draw <- decodeplot(plot)
+  plotObject <- .decodeJaspPlotObject(plot, returnGrob = FALSE, decodeContext = decodeContext)
+  plot2draw <- plotObject
```
vandenman: this should be reverted, see prior comments about encoding.

src/jaspPlot.cpp
```diff
@@ -107,8 +107,10 @@ void jaspPlot::renderPlot()
 			writeResult = tryToWriteImage(Rcpp::_["width"] = _width, Rcpp::_["height"] = _height, Rcpp::_["plot"] = obj, Rcpp::_["oldPlotInfo"] = oldPlotInfo, Rcpp::_["relativePathpng"] = R_NilValue);
 		}
 
-		// we need to overwrite plot functions with their recordedplot result
-		if(Rcpp::is<Rcpp::Function>(obj) && writeResult.containsElementNamed("obj"))
+		// Keep the state object in sync with the rendered object. R returns an
+		// editable decoded object for ggplot-like plots and a materialized object
+		// for function/base plots.
+		if(writeResult.containsElementNamed("obj"))
```
vandenman, line 113: why did this need to change? This was not broken?

R/resultDecoding.R
```diff
@@ -0,0 +1,343 @@
...
+  if (!is.character(x) || length(x) == 0L || length(columnMapping) == 0L)
+    return(x)
+
+  for (encoded in names(columnMapping))
+    x <- gsub(encoded, unname(columnMapping[[encoded]]), x, fixed = TRUE)
+
+  x
+}
+
+.decodeJaspFactorValues <- function(x, fieldName = NULL, decodeContext) {
```
vandenman, line 196: this reimplements the decoding from jasp. But is this really the best appraoch? Shouldn't we just reinstantiate that object? See https://github.com/jasp-stats/jasp-desktop/blob/development/Common/columnencoder.h, we could make an R6 object in jaspSyntax that provides an API to this and then use that instead? This is reimplementing the exact logic in that file.

vandenman (changes requested):

FBartos:
@vandenman I pushed the follow-up implementation based on the decoder-ownership discussion.

The current architecture is now split as follows:
- desktop/SyntaxInterface owns native ColumnEncoder snapshots and text-token replacement: jasp-desktop#6249
- jaspSyntax exposes the native decoder snapshot/text API to R: jaspSyntax#8
- jaspBase owns the R-facing decode boundary and decides which JASP-owned surfaces are decoded here

In this PR, jaspBase now carries the captured decoder context through result materialization, decodes tables/footnotes/stored figures/ggplot labels, and adds plot-specific decoding for ggplot data, mappings, layer data/mappings, facet mappings, and plot metadata. It still keeps arbitrary analysis-owned model state opaque, so this should not reintroduce the earlier too-deep object rewriting problem.

This also fixes the MixedModelsLMM smoke case that exposed encoded names inside the returned plot object: with patched jaspBase installed into a temporary R library, 
es() no longer has encoded JaspColumn_ / jaspColumn tokens in the broad result scan.

Validation on Windows/R 4.6: focused jaspBase tests, focused jaspSyntax tests after rebuilding/installing against local SyntaxInterface, SyntaxInterface export check 16/16, and the MixedModelsLMM smoke above.

FBartos:
Addressed the silent fallback concern in 3189fcc.\n\nWhat changed:\n- Removed .decodeJaspColumnTextWithMapping() and the R-side mapping/gsub fallback from result decoding.\n- Removed the llowLiveFallback path through decodeColNames().\n- Result/state/plot/condition decoding now propagates native jaspSyntax::decodeColumnText() failures instead of returning encoded text.\n- decodeColNames() still passes plain text through outside JASP, but now errors if an encoded bridge token is encountered without an installed decoder.\n- decodeName() no longer swallows decoder errors before passing names into C++ wrappers.\n\nThe intended ownership is now strict: jaspBase owns R result traversal only; token replacement must come from jaspSyntax -> native SyntaxInterface / ColumnEncoder. jaspSyntax PR #8 contains the matching bridge cleanup; jasp-desktop #6249 owns the native decoder API.\n\nFocused checks passed locally against source-loaded jaspSyntax and jaspBase:\n- 	est-result-object-decoding.R\n- 	est-runWrappedAnalysis.R\n\nConsumer checks in jaspTools #79 also passed with local source-loaded jaspSyntax/jaspBase:\n- 	est-rbridge-shim.R\n- 	est-jaspSyntax-lifecycle.R\n- 	est-generated-example-tests.R\n\nThe native runtime still emits the known stack-imbalance warnings at process shutdown in the runWrappedAnalysis-focused test, but assertions pass.

FBartos:
@vandenman I pushed the architecture-level follow-up from the decoding review.

jaspBase no longer receives or normalizes R-side column mappings. It carries the opaque jaspSyntax/SyntaxInterface encoder context and treats native decoding failures for encoded tokens as bugs, while still limiting eager decoding to JASP-owned result copies rather than downstream module internals.

Related PRs: jasp-desktop https://github.com/jasp-stats/jasp-desktop/pull/6249, jaspSyntax https://github.com/jasp-stats/jaspSyntax/pull/8, jaspTools https://github.com/jasp-stats/jaspTools/pull/79.

R/writeImage.R
FBartos: Agreed; the later architecture now handles both cases without relying on the live dataset. `runJaspResults()` captures a per-analysis decode context after dataset preload, `finishJaspResults()` / `.saveState()` materialize stored state with that context, and `state$figures` stores decoded plot objects for replay/export. For saved or reloaded results, `saveImage()`, `editImage()`, and `rewriteImages()` use the stored-result-state path instead of the currently active dataset. The actual token replacement now comes from the opaque jaspSyntax/SyntaxInterface encoder context rather than an R-side mapping.

R/resultDecoding.R
FBartos: Yes. This changed in the later fallback-removal pass. The R-side mapping/requestedMapping fallback is gone, and encoded-token decoding now goes through `jaspSyntax::decodeColumnText()` backed by native SyntaxInterface. If encoded legacy state is encountered without a usable context, jaspBase warns that it cannot safely decode it instead of silently applying a stale or approximate R mapping; native decoder failures for encoded tokens propagate as errors.

R/common.R
FBartos: Agreed. `saveImage()` now decodes through `.decodeJaspPlotObject(..., returnGrob = TRUE, decodeContext = .jaspDecodeContext(source = stored-result-state))`, so the saved/printed object uses the stored result context and avoids showing encoded labels from a live-dataset mismatch.

R/common.R
FBartos: This path changed with the stored-state architecture. We no longer rely on a live decoder or raw encoded object here; `rewriteImages()` decodes the stored plot object with `.jaspDecodeContext(source = stored-result-state)`, edits the decoded editable object, and then assigns that decoded object back to `plotObject`. The actual rendering path still runs through `writeImageJaspResults()`, which keeps state/rendered output synchronized.

R/common.R
FBartos: Same resolution as the prior plot-state comments: this is now using the stored-result-state context rather than live analysis state. `editImage()` decodes the stored editable plot object before resize/edit and assigns the decoded object back, while `writeImageJaspResults()` remains responsible for rendering and returning the object that should be persisted.

R/common.R
FBartos: Agreed, this belongs in Desktop rather than jaspBase. The corresponding Desktop PR now changes `Common/log.cpp` so `logType::null` is a real sink instead of `std::cout`, and the SyntaxInterface bridge can be silenced from jaspSyntax via the verbose toggle. See jasp-desktop#6249, specifically the `Common/log.cpp` reply explaining why `syntaxBridgeSetVerbose(false)` needed that change.

R/writeImage.R
FBartos: Addressed by moving the decode ownership rather than keeping the earlier R-side approach. `writeImageJaspResults()` now receives/passes a decode context, decodes the plot through jaspBase's JASP-owned result boundary, and delegates token replacement to native jaspSyntax/SyntaxInterface. The R-side mapping fallback that motivated this concern has been removed.

src/jaspPlot.cpp
FBartos: This became necessary once we eagerly store decoded plot state. Without writing the returned object back into `plotInfo[obj]`, the PNG could be rendered from a decoded/materialized object while the stored plot object remained encoded. The C++ change keeps the persisted state object in sync with the object returned by `writeImageJaspResults()` for all plot types, not only function/base plots.

R/resultDecoding.R
FBartos: Agreed. The current implementation no longer reimplements `ColumnEncoder` token replacement in jaspBase. That logic moved to the native bridge in jasp-desktop#6249, and jaspSyntax#8 carries the opaque Desktop encoder context plus calls native `decodeColumnText()`. jaspBase now only decides which JASP-owned result surfaces to traverse and passes encoded text to jaspSyntax/native decoding; the R mapping/gsub fallback has been removed.

R/common.R
FBartos: Small correction to the cross-PR note: the final Desktop implementation no longer changes `Common/log.cpp` or the global `logType::null` default. That broader approach was reverted after Joris' review.

The current Desktop shape is narrower: SyntaxInterface initializes logging with the same Boost null stream pattern used by Engine/MainWindow, then toggles `Log::setWhere()`/`Log::setDefaultDestination()` for the bridge. So this remains owned by the native SyntaxInterface/desktop layer, but does not change global logging behavior in Common.

boutinb (approved):
I have tested jaspSyntax with this jaspBase, and it works.
If @vandenman agrees, we could merge this.

R/resultDecoding.R
```diff
@@ -0,0 +1,316 @@
...
+        call. = FALSE
+      )
+    }
+  )
+  if (is.character(decoded) && length(decoded) == length(x)) {
+    names(decoded) <- names(x)
+    decoded
+  } else {
+    stop("Native jaspSyntax column decoder returned an invalid result.", call. = FALSE)
+  }
```
vandenman: Following the "parse don't validate" idea, just call the decoder immediately. Do not check for the presence of an encoded token, the columnencoder can do this much more efficiently. Do not validate the results afterward. All this should and is handled by the columnEncoder itself.

FBartos: Implemented in `1b414bf`.

`.decodeJaspColumnText()` now only unwraps the stored decode context and delegates directly to `jaspSyntax::decodeColumnText()` whenever a context is available. I removed the jaspBase-side encoded-token pre-scan and the post-call result-shape validation, so token/no-token handling and decoder result contracts stay below this layer.

I also added a focused regression test that verifies plain text is still delegated when a decode context is present; that would have failed with the old pre-scan.

Focused checks passed locally with the matching local jaspSyntax/SyntaxInterface installed first on `.libPaths()`:
- `test-result-object-decoding.R`
- `test-runWrappedAnalysis.R`

vandenman:
Changes look good but this needs jaspSyntax to be available/ installable first.

PR #211, opened 2026-07-20
RDS results export, plot `export` tag, and table type-inference fixes

RensDofferhoff:
Summary

This PR adds the ability to save analysis results as RDS alongside JSON, introduces a machine-readable `export` property on plots, and fixes numeric columns being misclassified as strings when they contain placeholder/missing values.

---

### 1. Save results as RDS (`jaspResults`)

- `saveResults()` now writes an `.rds` file next to the `.json` results file.
- When the `JASP_RDS_STRIP` environment variable is set (by JASP's `ProcessHelper`), bulky objects are stripped before saving:
  - `jaspObjectEnvironment` attributes (XPtrs to live C++ objects, ~90 MB bloat)
  - `plotObject` in `jaspPlotWrapper` lists (live ggplot objects with huge environments, ~57 MB each)
- New R helper in `R/resultsRdsHelpers.R` (`.jaspResults_stripEnv`) performs the recursive stripping.
- The `.rds` path is added to the keep-list so it isn't cleaned up.
- `saveResults()` is now called **before** `send()` in `complete()` so the RDS is available when the engine signals completion.

### 2. New `export` property on `jaspPlot`

- Adds a `Json::Value _export` field to `jaspPlot`, serialized in `dataEntry()`, `convertToJSON()`, and `convertFromJSON_SetFields()`.
- Exposed to R via `jaspPlot_Interface::setExport` / `getExport` and the `RCPP_MODULE` registration.
- The R6 wrapper `jaspPlotR` gets an `export` active binding:
  ```r
  plot$export <- list(medianDelta = 0.45, ciLow = 0.12, ciHigh = 0.78)
  ```
- Included in `toRObject()` as plain JSON (not an environment), so it **survives RDS stripping**.
- Purpose: lets analysis authors tag machine-readable data (effect sizes, credible intervals, Bayes factors at specific prior widths, etc.) onto plots for downstream consumers like RoboReport.

### 3. Fix: numeric columns becoming "string slop" due to missing values

- **`jaspTable::toRObject()`**: integer and numeric columns now check `cell.isNumeric()` before converting. Non-numeric cells (placeholders like `""`, `"."`, `"NaN"`, `"null"`) become `NA_INTEGER` / `NA_REAL` instead of calling `.asInt()` / `.asDouble()` on a string.
- **`jaspTable::deriveColumnType()`**: placeholder strings are skipped during type inference, so a column with e.g. `df = ""` for Mann-Whitney no longer gets classified as `composite`/`string`.

### 4. Fix: container `toRObject()` key

- `jaspContainer::toRObject()` now uses the child's `key` instead of `child->_title` when building the R list, fixing incorrect element naming.


RensDofferhoff:
Could one of you check this.
Need these changes for roboreport

JorisGoosen:
The code looks good, will test it in a bit.

One question though, when would we ever want the bulky objects included in the RDS? They are already stored as rds separately in `state`. So it would just be a duplicate. This means that the RDS strip thing can simply be the only behaviour.

Edit: Because im assuming this file is added to jaspfiles through the keeplist.

JorisGoosen:
Im having build issues unrelated to this PR but the code with the changes looks good to me.

JorisGoosen (approved):

